PR #24, opened 2022-02-10
Refactor: rename examples

leventebalogh:
This is really **just an idea** which thrown up in the air.

### Why?
When I first looked at the examples folder it was kind of hard to oversee what folders to panel plugins, which ones to datasource plugins, etc. Obviously I have skipped reading through the README (which has this information), but I guess this probably what most of other devs would do as well: **"show me the code!"**

Screenshot 2022-02-10 at 10 24 07

--- 

_It's probably not important and absolutely subjective, but still wanted to ask your opinions._


jackw (approved):
LGTM! 👍

mckn:
Love it! Awesome initiative <3

PR #45, opened 2022-02-16
Example: adding end-to-end tests for basic-datasource

mckn:


examples/datasource-basic/src/components/ConfigEditor/ConfigEditor.tsx
```diff
@@ -17,19 +17,25 @@ export function ConfigEditor(props: EditorProps): ReactElement {
...
         <SecretFormField
           tooltip="API Key used to make calls to your data source"
-          isConfigured={Boolean(secureJsonData?.apiKey)}
+          isConfigured={Boolean(secureJsonFields['apiKey'])}
           value={secureJsonData?.apiKey || ''}
           label="API Key"
           placeholder="secure json field (backend only)"
           labelWidth={6}
           inputWidth={20}
+          aria-label="Secret api key"
```
hugohaggmark, line 38: Super nit: we're trying to move things from aria-labels to data-testids. Unless the aria-label makes sense here from an a11y perspective. Also, instead of using this hardcoded string you should be using the same selector that you're using in the e2e tests.

examples/datasource-basic/cypress/integration/config.spec.ts
```diff
@@ -0,0 +1,34 @@
+import { e2e } from '@grafana/e2e';
+
+const configEditor = e2e.getSelectors({
+    timeField: 'Macro default time field',
+    apiKey: 'Secret api key',
+});
```
hugohaggmark, line 6: Nice but I would expect this to be used in the production code too, not just in the test code. I think there's were our thin layer upon Cypress shines.

hugohaggmark (approved):
LGTM 🎉

examples/datasource-basic/src/components/ConfigEditor/ConfigEditor.tsx
mckn: Great feedback! I started with that but when I used the `e2e.getSelectors` function it seems like it is expecting `aria-labels`.

examples/datasource-basic/cypress/integration/config.spec.ts
mckn: I will update!

examples/datasource-basic/src/components/ConfigEditor/ConfigEditor.tsx
```diff
@@ -17,19 +17,25 @@ export function ConfigEditor(props: EditorProps): ReactElement {
...
+            value={jsonData?.defaultTimeField ?? ''}
+          />
         </InlineField>
       </FieldSet>
 
       <FieldSet label="API Settings">
         <SecretFormField
           tooltip="API Key used to make calls to your data source"
-          isConfigured={Boolean(secureJsonData?.apiKey)}
+          isConfigured={Boolean(secureJsonFields['apiKey'])}
```
leventebalogh, line 32: **Nit:** I think we can use the dot notation here for accessing the property (I don't know if we have any standard for these)
```suggestion
          isConfigured={Boolean(secureJsonFields.apiKey)}
```

leventebalogh (approved):
Nice one 👏

PR #66, opened 2022-04-05
Chore: make the `basic-panel` example up to date with our plugin scaffolding

leventebalogh:
**Related issue:** https://github.com/grafana/grafana-plugin-examples/issues/65

**What changed?**
We have recently updated the plugin scaffolding generated by https://github.com/grafana/create-plugin to contain a few fixes and additions (e.g. E2E testing). This PR updates the Basic Panel example to have all those changes.


examples/panel-basic/package.json
```diff
@@ -47,11 +62,11 @@
   },
   "dependencies": {
     "@emotion/css": "^11.1.3",
-    "@grafana/data": "8.3.4",
-    "@grafana/runtime": "8.3.4",
+    "@grafana/data": "latest",
```
jackw, line 65: I think it would be best if we stick to "pinned version numbers" as before rather than the latest keyword. In this way we can see which version of the @grafana packages the plugin is using without needing to look in the yarn.lock file.

leventebalogh: Good point! 👍 

Fixed in [54fd23b](https://github.com/grafana/grafana-plugin-examples/pull/66/commits/54fd23b0213c5e7357bc59512f6848a794b8d9bb).

academo (approved):

PR #67, opened 2022-04-06
Chore: make the `basic-datasource` example up to date with our plugin scaffolding

leventebalogh:
**Related issue:** https://github.com/grafana/grafana-plugin-examples/issues/65

**What changed?**
We have recently updated the plugin scaffolding generated by https://github.com/grafana/create-plugin to contain a few fixes and additions (e.g. E2E testing). This PR updates the Basic Panel example to have all those changes.


examples/datasource-basic/package.json
```diff
@@ -47,10 +60,9 @@
   },
   "dependencies": {
     "@emotion/css": "^11.1.3",
-    "@grafana/data": "8.3.6",
-    "@grafana/runtime": "8.3.6",
-    "@grafana/toolkit": "8.3.6",
-    "@grafana/ui": "8.3.6",
+    "@grafana/data": "latest",
```
jackw, line 63: I think it would be best if we stick to "pinned version numbers" as before rather than the latest keyword. In this way we can see which version of the `@grafana` packages the plugin is using without needing to look in the `yarn.lock` file.

leventebalogh: Good point! 👍 

Fixed in [2b03ee4](https://github.com/grafana/grafana-plugin-examples/pull/67/commits/2b03ee40e150c2993a31db9bafab2abcabcedfe3).

jackw (approved):
🚀

PR #70, opened 2022-04-28
Basic App: update the `plugin.json`

leventebalogh:
### What changed?
- **Updated the routes** (under `includes`) **in the `plugin.json`** to match with the current routing
- **Updated the `{ defaultNav }` properties for the routes** to only have a single default route (otherwise it would pick the last `{ defaultNav: true } as a default route.


jackw (approved):
LGTM! 🚀

PR #75, opened 2022-07-19
Fix e2e tests for datasource-basic example and remove cypress docker

academo:
* Removes the usage of docker to run cypress
* Adds retries to cypress for flaky test
* Sets a timezone to run cypress to prevent local timezone issues.

### DIFF
--- a/examples/datasource-basic/cypress.json
+++ b/examples/datasource-basic/cypress.json
@@ -1,4 +1,7 @@
 {
-    "video": false,
-    "reporter": "spec"
+  "video": false,
+  "reporter": "spec",
+  "retries": {
+    "runMode": 2
+  }
 }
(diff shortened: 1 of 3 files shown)


examples/datasource-basic/cypress.json
```diff
@@ -1,4 +1,7 @@
 {
-    "video": false,
-    "reporter": "spec"
-}
\ No newline at end of file
+  "video": false,
+  "reporter": "spec",
+  "retries": {
+    "runMode": 2
```
leventebalogh, line 5: **Question:** was this needed because it was flaky otherwise, or it was just added for "safety"?

leventebalogh (approved):
LGTM 🚀

examples/datasource-basic/cypress.json
academo: Yes. The test was flaky without apparent reason and most of the time re-running it fixes this flakyness.

PR #76, opened 2022-07-19
Remove e2e tests from app-basic.

academo:
The app basic plugin doesn't have e2e tests and it is wrongly triggering a github pipeline.


leventebalogh (approved):
I think it is fine for now, maybe (hopefully) we introduce some basic E2E tests in the future.

PR #84, opened 2022-09-01
Migrate panel examples

academo:
All the `panel-*` examples use the same configuration with only their internal components code changing.

This PR migrates them all at once to use the create tool.

Additionally a new workflow is added so pull requests check that the plugin examples can be built in CI.

Note: all the migrations were tested locally in grafana 9.1.2 and confirmed to work.


mckn (approved):
LGTM! Great work!

Issue #53, opened 2022-03-04
Basic App Plugin: add integration tests


leventebalogh:
Add integration tests that verify that the features that being used from Grafana plugin API are working as expected.

Issue #12, opened 2022-01-20
Examples: app with several pages


mckn:
**Related pull request(s):** https://github.com/grafana/grafana-plugin-examples/pull/26

### Description
Create a plugin that contains an app with several pages. This should be an example of how to create a basic app without any advance functionality. 

### Tasks
- [x] Root page with routing (**Done in** https://github.com/grafana/grafana-plugin-examples/pull/26)
   - [x] Showcase how create page with tabs.
   - [x] Showcase how create page without tabs.
   - [x] Showcase how create full width page.
   - [x] Showcase how to navigate between pages.
- [x] Configuration page. (**Done in** https://github.com/grafana/grafana-plugin-examples/pull/26)
   - [x] Showcase how to manage secrets e.g. API keys etc.
   - [x] Showcase how to use the form components.
- [x] #53 (**Done in** https://github.com/grafana/grafana-plugin-examples/pull/92)

PR #92, opened 2022-09-15
Example: adding tests to verify the basic app

mckn:
This should probably be good enough to verify that this app is working as expected. I renamed the test specs in the datasource to be consistent with the rest of the spec files in the repo.

Fixes: #53 #12


examples/app-basic/src/plugin.json
```diff
@@ -2,7 +2,7 @@
   "$schema": "https://raw.githubusercontent.com/grafana/grafana/master/docs/sources/developers/plugins/plugin.schema.json",
   "type": "app",
   "name": "Basic App",
-  "id": "basic-app",
+  "id": "grafana-basic-app",
```
mckn, line 5: Prefixing with `grafana-` to remove risk of colliding.

examples/app-basic/cypress/integration/01-app-config.spec.ts
```diff
@@ -0,0 +1,26 @@
+import { e2e } from '@grafana/e2e';
+import { testIds } from '../../src/components/testIds';
+
+const { appConfig } = e2e.getSelectors(testIds);
+
+describe('configurating app', () => {
+  beforeEach(() => {
+    cy.visit('http://localhost:3000/plugins/grafana-basic-app');
```
jackw: We should be able to improve this with a flow to e2e package or some such.

academo (approved):
Nice! thanks for adding those tests.

examples/app-basic/cypress/integration/02-app-navigation.spec.ts
```diff
@@ -0,0 +1,51 @@
...
+  });
+
+  it('should render full width example successfully', () => {
+    // wait for page to successfully render
+    pageOne.container().should('be.visible');
+    cy.get('h1').contains(headerTitle).should('be.visible');
+    
+    // navigating to page four with full width layout
+    pageOne.navigateToFour().click();
+    cy.get('body').find('h1', ).should('not.exist');
```
jackw, line 26: Is this syntax correct?

```suggestion
    cy.get('body').find('h1').should('not.exist');
```

examples/app-basic/src/plugin.json
jackw: Isn't the prefix of a plugin id be the org name? Maybe swap it out for `myOrg`?

`"id": "myorg-basic-app",`

jackw (approved):
Awesome work @mckn! 🚀 

Left a couple of nits.

examples/app-basic/cypress/integration/01-app-config.spec.ts
mckn: I totally agree! Lets add it to the list to expose some kind of function that will fetch this based on the environment we are running the tests in.

examples/app-basic/cypress/integration/02-app-navigation.spec.ts
mckn: Oops, my bad 🙈  Thanks for the ping.

PR #96, opened 2022-09-21
Build: Changed so we use plugin id as container name

mckn:
This PR will use the plugin id as the container name when running integration tests. I have a feeling that now and then (hard to reproduce) we manage to kill a container belonging to another plugin example. This will prevent that from happening.


jackw (approved):
LGTM! 🚀

PR #79, opened 2022-08-25
Panels: Introduce example using datalinks

jackw:
This PR introduces a panel plugin example that renders datalinks. 

Screenshot 2022-08-25 at 16 50 25

*Notes for reviewers*
- Nearly all the files in this PR are scaffolding for a new plugin. [Filtering files by .tsx](https://github.com/grafana/grafana-plugin-examples/pull/79/files?file-filters%5B%5D=.tsx&show-viewed-files=true) might help with a review.
- I've not attempted to add datalinks to a plugin before so not 100% certain this panel is doing it the preferred way hence why I've opened this PR as a draft.

PR #97, opened 2022-09-23
Fix datalinks failing tests

jackw:
This is a follow up to #79 which was merged with failing tests. This PR does the following:

- [x] Updates the datalinks example to make it compatible with Grafana 9.0+
- [x] Fixes the plugin name in artefact names on a test failure
- [x] Updates datalinks test to remove duplicate test scenarios
- [x] Fixes failing datalinks test by introducing a cypress.json to disable video recording (we believe grafana/grafana/pull/53342 is too much for gh actions and causes tests to hang when trying to encode video to mp4)


examples/panel-datalinks/src/components/DataLinksPanel.tsx
```diff
@@ -76,3 +78,14 @@ const getStyles = () => {
     `,
   };
 };
+
+// This one is created to support versions prior to 9.1.0
```
leventebalogh, line 82: Sorry, I am not sure I fully understand - is it because `config` was part / not part of the props of `DataLinksContextMenu` previously?

leventebalogh (approved):
LGTM 👍

examples/panel-datalinks/src/components/DataLinksPanel.tsx
mckn: It is required in 9.0 but dropped in 9.1

leventebalogh: I see, thanks 🙏

mckn: So maybe we need to improve that comment?

leventebalogh: Hmm, yeah, probably that would make it easier to understand. Maybe something like the following?
```suggestion
// The `config` prop was required in <9.1.0 versions but got dropped in 9.1.0. This type is here to support both Grafana version ranges.
```

.github/workflows/integration-tests.yml
```diff
@@ -29,7 +29,7 @@ jobs:
       - name: Set the name of the plugin-example to be tested
         id: example-name
         run: |
-          basename ${{ matrix.pluginDir }}
+          echo "::set-output name=PLUGIN_NAME::$(basename ${{ matrix.pluginDir }})"
```
academo, line 32: nice. Thanks!

examples/panel-datalinks/src/plugin.json
```diff
@@ -8,7 +8,10 @@
     "author": {
       "name": "My org"
     },
-    "keywords": ["panel", "datalinks"],
+    "keywords": [
```
academo, line 11: @mckn there's something in your editor configuration not picking up the prettier configuration.  I notice in this (and other files) there are some changes in quotes and spaces.

In my case if I run prettier in this file I get `"keywords": ["panel", "datalinks"],` (all in one line) vs here in separated lines. 

Could you verify what is the result if you run prettier in CLI against this file?

jackw: I too get the same result as @academo when saving the plugin.json. I've gone ahead and updated example files with `prettier -w`.

academo: Somehow I got confused and thought @mckn was the PR author. Thanks for checking!

PR #98, opened 2022-09-29
Basic App: use the new `PluginPage` component in a backwards compatible way.

leventebalogh:
**This PR is an experiment.** It is mostly based on the logic introduced in https://github.com/grafana/grafana-starter-app/pull/39.

## What changed?
We have introduced a new `PluginPage` component in https://github.com/grafana/grafana/pull/53174, however unfortunately it's not straightforward to make it work in older versions of Grafana (<9.2). My intention was to try to give a "oneliner" solution for plugin developers to benefit from the new functionality while still being compatible with older versions of Grafana, and for this I shifted the migration-related logic into a separate package which we would own (experiment). 

[Checkt the experimental package `@grafana/migrations` &rarr;](https://github.com/grafana/migrations/tree/main/packages/migrations-9.2)

|  **New Top Navigation**  |    **Fallback (older versions)**    |
|--------------|-----------------|
| Screenshot 2022-09-29 at 12 09 25 | Screenshot 2022-09-29 at 12 10 55 |

### DIFF
--- a/examples/app-basic/docker-compose.yaml
+++ b/examples/app-basic/docker-compose.yaml
@@ -9,6 +9,8 @@ services:
         grafana_version: ${GRAFANA_VERSION:-9.1.2}
     ports:
       - 3000:3000/tcp
+    environment:
+      - GF_FEATURE_TOGGLES_TOPNAV=true
     volumes:
       - ./dist:/var/lib/grafana/plugins/app-basic,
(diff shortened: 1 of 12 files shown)


leventebalogh:
We have decided (Plugins Platform daily, 2022-10-03) that we won't integrate this feature into our plugin examples until the top-nav is enabled even without a feature flag.

PR #99, opened 2022-10-21
Feat: Add app plugin example with dashboards

### LINKED ISSUE #54: App: add an example with dashboards
### What we would like to achieve?
We would like to have a separate example that shows how to ship and register dashboards using an app plugin.
We have decided with @jackw that we would create a separate example for this rather than including it into the `Basic App` example, basically to keep the examples more "single purpose" and easier to reason about.

tolzhabayev:
This PR adds an app plugin example with dashboards.

Skeleton is generated using [create-plugin](https://github.com/grafana/plugin-tools/tree/main/packages/create-plugin).

**Additional changes**
- bumping node to 14
- fixing plugin id's to fit our schema
- small changes in provisioning of the app-basic example
- bumping dependencies in github actions to reduce deprecation warning

Fixes: #54


examples/app-basic/docker-compose.yaml
```diff
@@ -10,5 +10,5 @@ services:
     ports:
       - 3000:3000/tcp
     volumes:
-      - ./dist:/var/lib/grafana/plugins/app-basic,
```
tolzhabayev, line 13: I guess this comma was a typo

.github/workflows/integration-tests.yml
```diff
@@ -7,13 +7,13 @@ jobs:
     outputs:
       pluginDirs: ${{ steps.set-plugin-dirs.outputs.pluginDirs }}
     steps:
-      - uses: actions/checkout@v2
```
tolzhabayev, line 10: bumping all the actions to v3 to use node 16

tolzhabayev:
@leventebalogh should be ready for review now

examples/app-basic/package.json
```diff
@@ -11,14 +11,14 @@
     "typecheck": "tsc --noEmit",
     "lint": "eslint --cache --ignore-path ./.gitignore --ext .js,.jsx,.ts,.tsx .",
     "lint:fix": "yarn lint --fix",
-    "server": "docker-compose up -d --build"
```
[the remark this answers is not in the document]
academo: Yes it is better without the detached because it is clear for the dev that something is running in a console.

examples/app-with-dashboards/src/module.ts
```diff
@@ -0,0 +1,12 @@
+import { AppPlugin } from '@grafana/data';
+import { App } from './components/App';
+import { AppConfig } from './components/AppConfig';
+
+export const plugin = new AppPlugin<{}>().setRootPage(App).addConfigPage({
+  title: 'Configuration',
+  icon: 'cog',
+  // @ts-ignore - Would expect a Class component, however works absolutely fine with a functional one
```
academo, line 8: I prefer we don't ship examples with typescript ignore errors.

tolzhabayev: Good catch, we have to update our templates in create-plugin.

I am not sure why it was there, maybe it was needed previously. It seems to build properly without it. 
https://github.com/grafana/grafana-plugin-examples/pull/99/commits/8884b24ee50f3bffa928d463ec1a1447457178f6

and 

https://github.com/grafana/grafana-plugin-examples/pull/99/commits/1e13559b4b5f08c078c117c77d064ca8928899f3

tolzhabayev: https://github.com/grafana/plugin-tools/pull/124

academo (approved):
Nice one!

PR #103, opened 2022-10-28
Feat: Add http backend example plugin

xnyo:
This PR adds a datasource example where data is fetched from a remote API, from the backend.

It is similar to [examples/datasource-http](https://github.com/grafana/grafana-plugin-examples/tree/main/examples/datasource-http), but the data fetching happens on the backend.

This allows to use alerting with the data source.

An example HTTP server that returns data in a compatible format is included in `cmd/server`, and it must be started separately. (build with `mage server`, then run with `./cmd/server/server`).

Tests are included for both the data source plugin and for the example server.

No querying capabilities are supported for simplicity's sake (the values from the remote HTTP API are returned as-is as a dataframe).


examples/datasource-http-backend/cypress/integration/01-smoke.spec.ts
```diff
@@ -0,0 +1,10 @@
+import { e2e } from '@grafana/e2e';
+
+e2e.scenario({
```
marefr, line 3: Guess this was added by create-plugin? Probably a discussion regarding create-plugin, but leaving it here so we don't forget about it. I guess we would like to invite plugin developers to create integration tests and by that provide some boilerplate if it's a panel, a datasource etc?

examples/datasource-http-backend/pkg/plugin/datasource.go
```diff
@@ -0,0 +1,150 @@
...
+
+	httpClient *http.Client
+}
+
+// Dispose here tells plugin SDK that plugin wants to clean up resources when a new instance
+// created. As soon as datasource settings change detected by SDK old datasource instance will
+// be disposed and a new one will be created using NewSampleDatasource factory function.
+func (d *Datasource) Dispose() {
+	// Clean up datasource instance resources.
+	d.httpClient.CloseIdleConnections()
```
marefr, line 56: Nice. Wasn't aware of this :+1:

examples/datasource-http-backend/pkg/plugin/datasource.go
```diff
@@ -0,0 +1,150 @@
...
+func (d *Datasource) query(ctx context.Context, pCtx backend.PluginContext, query backend.DataQuery) (backend.DataResponse, error) {
+	// Response to be returned.
+	var response backend.DataResponse
+
+	// Do HTTP request
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, d.settings.URL, nil)
+	if err != nil {
+		return response, fmt.Errorf("new request with context: %w", err)
+	}
+	resp, err := d.httpClient.Do(req)
```
marefr, line 90: Should always close the response body. Running `mage lint` doesn't complain about this?

examples/datasource-http-backend/pkg/plugin/datasource.go
```diff
@@ -0,0 +1,150 @@
...
+}
+
+// CheckHealth performs a request to the specified data source and returns an error if the HTTP handler did not return
+// a 200 OK response.
+func (d *Datasource) CheckHealth(ctx context.Context, _ *backend.CheckHealthRequest) (*backend.CheckHealthResult, error) {
+	r, err := http.NewRequestWithContext(ctx, http.MethodGet, d.settings.URL, nil)
+	if err != nil {
+		return newHealthCheckErrorf("could not create request"), nil
+	}
+	resp, err := d.httpClient.Do(r)
```
marefr, line 133: Should always close the response body. Running `mage lint` doesn't complain about this?

examples/datasource-http-backend/src/plugin.json
```diff
@@ -0,0 +1,38 @@
...
+    "links": [
+      {
+        "name": "Website",
+        "url": "https://github.com/grafana/grafana-starter-datasource"
+      },
+      {
+        "name": "License",
+        "url": "https://github.com/grafana/grafana-starter-datasource/blob/master/LICENSE"
+      }
+    ],
```
marefr, line 29: Probably also related to create-plugin. After reviewing quite a few plugins I often notice that these same links are not changed.

examples/datasource-http-backend/pkg/plugin/datasource.go
xnyo: My bad!

examples/datasource-http-backend/cypress/integration/01-smoke.spec.ts
xnyo: Yes. This was added by create-plugin. I can add some E2E test examples specific for this plugin

examples/datasource-http-backend/src/plugin.json
xnyo: Yes. This was added by create-plugin too

examples/datasource-http-backend/cypress/integration/01-smoke.spec.ts
marefr: Not sure. Let's ask the rest of the team what they think, like do we have E2E for the other examples etc.

examples/datasource-http-backend/src/plugin.json
marefr: I created https://github.com/grafana/plugin-tools/issues/129

examples/datasource-http-backend/pkg/plugin/datasource.go
```diff
@@ -0,0 +1,152 @@
...
+	// Do HTTP request
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, d.settings.URL, nil)
+	if err != nil {
+		return response, fmt.Errorf("new request with context: %w", err)
+	}
+	resp, err := d.httpClient.Do(req)
+	if err != nil {
+		return response, fmt.Errorf("http client do: %w", err)
+	}
+	defer resp.Body.Close()
```
marefr, line 94: Great. But would also like you to handle the case if `Body.Close` returns an error and log that. Otherwise those errors just being swallowed/ignored which is not great.

examples/datasource-http-backend/pkg/plugin/datasource.go
```diff
@@ -0,0 +1,152 @@
...
+func (d *Datasource) CheckHealth(ctx context.Context, _ *backend.CheckHealthRequest) (*backend.CheckHealthResult, error) {
+	r, err := http.NewRequestWithContext(ctx, http.MethodGet, d.settings.URL, nil)
+	if err != nil {
+		return newHealthCheckErrorf("could not create request"), nil
+	}
+	resp, err := d.httpClient.Do(r)
+	if err != nil {
+		return newHealthCheckErrorf("request error"), nil
+	}
+	defer resp.Body.Close()
```
marefr, line 138: Great. But would also like you to handle the case if `Body.Close` returns an error and log that. Otherwise those errors just being swallowed/ignored which is not great.

marefr (changes requested):
Great, almost there. Just a minor thing left.

marefr (approved):
LGTM great work

wbrowne (approved):
Took it for a spin and it looks great. Awesome work 🚀

PR #110, opened 2022-11-15
Fix: HTTP backend example: Removed double defer resp.Body.Close()

xnyo:
In the http backend example, `resp.Body.Close()` was accidentally called twice in `query`


marefr (approved):
LGTM nice

PR #112, opened 2022-11-15
Chore: schedule to run integration tests regularly

tolzhabayev:
This PR introduces scheduled runs once a day.
It will also notify slack when it runs.

I have to adjust the message format and also that it should only notify on failure later on once I've tested this enough.


jackw (approved):
Nice! 🚀

leventebalogh (approved):
Cool! 👏 

Do we always want to receive notifications about successful builds as well though? (I think we only need to action on failures?)

tolzhabayev:
> Cool! 👏
> 
> Do we always want to receive notifications about successful builds as well though? (I think we only need to action on failures?)

absolutely not, I am just calibrating the message / how it works and will keep it in there so that it always triggers and then remove the successful part so that it only notifies on failures.

tolzhabayev:
ok should be ready for review

.github/workflows/integration-tests.yml
```diff
@@ -111,42 +117,43 @@ jobs:
 
       ## Latest Version Tests
       ## Runs the plugin tests against the latest stable version of Grafana.
-      - name: Build plugin with latest version of e2e libraries
-        run: |
-          yarn upgrade @grafana/e2e-selectors @grafana/e2e --latest
-        working-directory: ${{ matrix.pluginDir }}
+      ## Activate this tests again once grafana/grafana sorts out the issue with npm packages mismatching the docker image tags
```
jackw, line 120: Can we uncomment this and try running it again please? Turns out the npm latest dist-tag was pointing at the 9.3.0-beta1. 🤦

tolzhabayev: Uncommented https://github.com/grafana/grafana-plugin-examples/pull/112/commits/5022ba636ed645041e6438a3540f145d1adab445

jackw (approved):
LGTM! 🚀 

Amazing work on getting this together @tolzhabayev ! 🥇

Issue #109, opened 2022-11-11
Add query data error handling to datasource-http example


tolzhabayev:
After merging https://github.com/grafana/grafana-plugin-sdk-go/pull/544 we should include query data error handling in datasource examples. Similar to https://github.com/grafana/grafana-plugin-examples/issues/108

**An example of a core plugin that needs to improve it's error handling to work properly with 207 responses:**
- https://github.com/grafana/grafana/blob/main/public/app/features/plugins/sql/datasource/SqlDatasource.ts#L167
- Currently the "save & test" will be "green" / "ok" even if the database responds with an error https://github.com/grafana/grafana/discussions/48975

leventebalogh:
@xnyo do you think this is closable? (as https://github.com/grafana/grafana-plugin-examples/pull/119 is merged)

PR #119, opened 2022-12-16
Datasource-http: Improve error handling and error messages

xnyo:
Improves error handling and error messages in the datasource-http example.
No 207-related changes to do, because the datasource does not have a backend.

Related to #109

### DIFF
--- a/examples/datasource-http/src/DataSource.ts
+++ b/examples/datasource-http/src/DataSource.ts
@@ -41,11 +41,20 @@ export class DataSource extends DataSourceApi<MyQuery, MyDataSourceOptions> {
        * }
        */
       const datapoints = response.data.datapoints;
+      if (typeof datapoints === 'undefined') {
+        throw new Error('Remote endpoint reponse does not contain "datapoints" property.');
+      }
 
       const timestamps: number[] = [];
       const values: number[] = [];
 
       for (let i = 0; i < datapoints.length; i++) {
+        if (typeof datapoints[i].Time === 'undefined') {
+          throw new Error(`Data point ${i} does not contain "Time" property`);
+        }
+        if (typeof datapoints[i].Value === 'undefined') {
+          throw new Error(`Data point ${i} does not contain "Value" property`);
+        }
         timestamps.push(datapoints[i].Time);
         values.push(datapoints[i].Value);
       }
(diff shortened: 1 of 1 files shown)


examples/datasource-http/src/DataSource.ts
```diff
@@ -41,11 +41,20 @@ export class DataSource extends DataSourceApi<MyQuery, MyDataSourceOptions> {
        * }
        */
       const datapoints = response.data.datapoints;
+      if (typeof datapoints === 'undefined') {
```
academo, line 44: use 

```
if (datapoints === undefined) {
```

instead.

small explanation: `typeof [var] === 'undefined'` is a bit less performant and it is only useful when you are not sure if the variable will be initialized or not. In this case you are sure the variable is initialized (in the previous line) and you can use the small optimization of comparing directly to undefined vs getting the type and comparing strings.

examples/datasource-http/src/DataSource.ts
```diff
@@ -41,11 +41,20 @@ export class DataSource extends DataSourceApi<MyQuery, MyDataSourceOptions> {
...
+        throw new Error('Remote endpoint reponse does not contain "datapoints" property.');
+      }
 
       const timestamps: number[] = [];
       const values: number[] = [];
 
       for (let i = 0; i < datapoints.length; i++) {
+        if (typeof datapoints[i].Time === 'undefined') {
```
academo, line 52: In this specific case you could argue, that `Time` might not be initialized but you are actually accessing a property of an already initialized variable `datapoints[i]` that might not exist and thus return undefined.

examples/datasource-http/src/DataSource.ts
```diff
@@ -93,7 +102,7 @@ export class DataSource extends DataSourceApi<MyQuery, MyDataSourceOptions> {
       if (_.isString(err)) {
         message = err;
       } else if (isFetchError(err)) {
-        message += err.statusText ? err.statusText : defaultErrorMessage;
+        message = `Fetch error: ${err.statusText ? err.statusText : defaultErrorMessage}`;
```
academo, line 105: Nit: re-write this to make it more readable, one might think this ternary condition will always return true if you don't look carefully:


```
message = "Fetch error: " + err.statusText ? err.statusText : defaultErrorMessage;
```

academo (approved):
I added a few nits to make the code neater but good to go if you want to skip them.

xnyo:
Thanks for the review @academo , I have applied your suggestions :smiley:

PR #129, opened 2023-02-20
Chore: Node 18 compatibility

jackw:
This PR contains the output from running `npx @grafana/create-plugin update` across all plugin examples for node 18 compatibility. See [create-plugin PR](https://github.com/grafana/plugin-tools/pull/144).


andresmgot:
It seems that ['resolve-as-bin'](https://github.com/grafana/grafana/blob/main/packages/grafana-e2e/cli.js#L28) is not compatible with Node 18 :sweat_smile:

jackw:
> It seems that ['resolve-as-bin'](https://github.com/grafana/grafana/blob/main/packages/grafana-e2e/cli.js#L28) is not compatible with Node 18 😅 ([last updated 4 years ago](https://www.npmjs.com/package/resolve-as-bin))

Yep, [I'm fixing that now](https://github.com/grafana/grafana/issues/59425). I think it best I close this and open a new PR that pins the version of node and npm in this repos workflow for the time being.

PR #135, opened 2023-03-14
Example: Show case how to extend Grafana core UI from plugins

mckn:
This PR introduces a new example to the repo showcasing how to extend the Grafana application UI from an app plugin.


examples/app-with-extensions/provisioning/plugins/app.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: 1
+
+apps:
+  - type: myorg-extensions-app
+    org_id: 1
+    org_name: Main Org.
+    disabled: false
```
jackw, line 7: Is this the provisioning equivalent of `enabled: true`? 🤔

jackw (approved):
LGTM. 🚀 

nit: I think it might be a good idea to highlight somewhere that this feature is only available in Grafana 9.5.0+ to save devs from scratching their heads if they try to implement this now.

examples/app-with-extensions/provisioning/plugins/app.yaml
mckn: Yes - so weird

PR #136, opened 2023-03-16
Add backend tracing example

xnyo:
Adds a plugin example to demonstrate how to use distributed tracing using Grafana's SDK.

If you want to try it out:

- Build it against this version of the SDK: https://github.com/grafana/grafana-plugin-sdk-go/pull/620
- And test it with this version of Grafana: https://github.com/grafana/grafana/pull/63714

If you just want to see how the API looks like, take a look at:
- pkg/plugin/datasource.go


marefr:
Wonder if we potentially could incorporate this example in https://github.com/grafana/grafana-plugin-examples/tree/main/examples/datasource-http-backend. I'm not sure what make most sense, many examples with specific examples or fewer with mixed examples. WDYT?

xnyo:
@marefr Good point. I have added tracing to the http-backend example as well, to see how it would work with a more "complex" plugin and I have just created a draft PR for this (see https://github.com/grafana/grafana-plugin-examples/pull/140). I think it would make sense to add tracing to the http-backend example because it would also showcase the http client middleware.

As a first-time plugin developer, I believe I would prefer having one "bigger" example with all the best practices and features, and eventually get rid of the stuff I don't want, rather than having to go back and forth between multiple examples. So honestly I would close this PR and add tracing to http-backend directly.

@tolzhabayev @sympatheticmoose WDYT?

tolzhabayev:
> @marefr Good point. I have added tracing to the http-backend example as well, to see how it would work with a more "complex" plugin and I have just created a draft PR for this (see #140). I think it would make sense to add tracing to the http-backend example because it would also showcase the http client middleware.
> 
> As a first-time plugin developer, I believe I would prefer having one "bigger" example with all the best practices and features, and eventually get rid of the stuff I don't want, rather than having to go back and forth between multiple examples. So honestly I would close this PR and add tracing to http-backend directly.
> 
> @tolzhabayev @sympatheticmoose WDYT?

Moving this example to the `http-backend` seems reasonable to me, I would love us to encourage more plugin devs to have tracing - it will make our operations and operations of anyone running plugins easier.

xnyo:
Closing this in favour of #140 (adding tracing to datasource-http-backend example)

PR #140, opened 2023-03-28
Add tracing to datasource-http-backend

xnyo:
Updates the datasource-http-backend example to showcase how to plugins developers can create custom spans from plugins.


examples/datasource-http-backend/pkg/plugin/datasource.go
```diff
@@ -39,6 +42,12 @@ func NewDatasource(settings backend.DataSourceInstanceSettings) (instancemgmt.In
 	if err != nil {
 		return nil, fmt.Errorf("http client options: %w", err)
 	}
+	// http clients created with httpclient.New have a middleware stack consisting
+	// of useful generic middlewares such as:
+	//  - TracingMiddleware (spans creates for each outgoing HTTP request)
```
andresmgot, line 47: nit: indentation here (and maybe move the TracingMiddleware to the last, since it's the most recent)

andresmgot (approved):
LGTM!

examples/datasource-http-backend/README.md
```diff
@@ -1,10 +1,18 @@
 # Grafana Data Source Plugin Template
 
-This example queries data from an HTTP API.
+This example queries data from an HTTP API, which returns data in JSON format and then it's converted to data frames.
 
-This differs from datasource-http, because the data fetching happens on the backend.
+This differs from the datasource-http example, because the data fetching happens on the backend.
```
marefr, line 5: Maybe something along these lines?

```suggestion
This differs from the datasource-http example, because the data fetching happens in the plugin backend rather than going through Grafana's datasource HTTP proxy.
```

marefr (commented):
Great. Added some suggestions

examples/datasource-http-backend/pkg/plugin/datasource.go
xnyo: I think those highlighted by @marefr are important to have documented somewhere, as it took me a while to understand what they did the first time I saw them, without reading the source code. We can add it to the example but maybe we should also have it in the docs somewhere?

Maybe we can have a new page under "Backend plugins" explaining how to use HTTP clients on backend plugins? However I believe I already saw something similar a while ago (maybe one of the plugin authentication docs), but I'm not sure.

WDYT?

andresmgot: the more documentation, the better :shrug:

xnyo: I'd say we can add what Marcus suggested to this plugin example for now to have it somewhere and then see if it's worth having it in the docs as well

marefr (approved):
LGTM

PR #142, opened 2023-04-05
Examples: Updated example to work with latest version of plugin extenions

mckn:
We have a dependency on the canary tag due to `9.5.0` not yet published to npm


leventebalogh (approved):
LGTM 👍

PR #147, opened 2023-04-19
Build: detect canary versions once per workflow

mckn:
Instead of detecting the `npm canary version` and `docker canary tag` in each test we do it once per workflow run. This will speed up the build but also make it fail fast in case docker hub is down or the canary docker image is missing.


jackw (approved):
Nice! Shaves ~1min off the integration test workflow run from what can tell. 🚀

PR #154, opened 2023-05-04
Example: Update example with show casing how to handle multi queries

mckn:
I'm updating the UI extension example to show case how to conditionally display a modal depending on number of queries.


examples/app-with-extensions/src/components/Modal/Modal.tsx
```diff
@@ -1,8 +1,10 @@
+import { DataQuery } from '@grafana/data';
 import { testIds } from 'components/testIds';
 import React, { ReactElement } from 'react';
 
 type Props = {
   panelTitle: string | undefined;
+  targets: DataQuery[] | undefined;
```
leventebalogh, line 7: Maybe it would make sense to somehow (in a dummy way) print these out in the modal body?

examples/app-with-extensions/src/module.tsx
```diff
@@ -43,17 +43,33 @@ export const plugin = new AppPlugin<{}>()
...
-        body: () => <Modal panelTitle={context?.title} />,
-      });
+      const targets = context?.targets;
+      const title = context?.title;
+
+      if (!Array.isArray(targets) || targets.length === 0) {
+        return;
+      }
+
+      if (targets.length > 1) {
```
leventebalogh, line 53: Maybe add a comment above this block to explain our intention?

```typescript
// Show a modal to display a UI for selecting between the available queries (targets) 
// in case there are more available. 
if (targets.length > 1) {
```

leventebalogh (approved):
LGTM 👍

PR #161, opened 2023-05-16
Update App Basic and App with Backend

jackw:
This PR updates the app basic example with the following:

- replaces `locationService.reload` with `window.location.reload`. `locationService.reload` doesn't appear to refresh the page
- replaces `SecretInput` component in the example with the component found in `grafana/ui`
- replaces usage of deprecated `backendSrv.datasourceRequest` with `backendSrv.fetch`
- updates e2e test now window reloads to check plugin settings are persisted

It also updates the app with backend example with the following:
- replaces `locationService.reload` with `window.location.reload`. `locationService.reload` doesn't appear to refresh the page
- replaces usage of deprecated `backendSrv.datasourceRequest` with `backendSrv.fetch`


leventebalogh (approved):
Nice one! 👏

PR #165, opened 2023-05-24
CI: Clean up GH actions

jackw:
This PR deletes the build-all workflow, only sets up go if backend exists, sets the go cache path to stop all the warnings from appearing in action runs and limits slack notifications to only happen on a scheduled run.

**Why delete the build all workflow?**
Currently PRs trigger two workflows `Build all plugins` and `Run integration tests`. This seems a bit pointless as `Run integration tests` has to build all plugins before an e2e test can be run.


leventebalogh (approved):
Nice 👏

Issue #167, opened 2023-05-25
Chore: Migrate examples to use NPM


jackw:
Yarn 1 is deprecated and nearing end of life. Most bugs that are present in the current yarn 1 version were "fixed" in yarn 2+. 

We could move to yarn berry but without enabling PnP I'm not sure it will bring us any benefits over NPM and as this repo contains multiple NPM projects (rather than a monorepo) it will mean each example gets a copy of yarn berry which we need to maintain. PNPM could be an alternative but it would likely require more time setting various hoist patterns than moving to NPM.

NPM however is available with Node and the migration should be relatively painless (`npm install --legacy-peer-deps`).

PR #168, opened 2023-05-25
Chore: Migrate to NPM

jackw:
Yarn 1 is deprecated and nearing end of life. This PR does the following in each example directory:

- Run `npx @grafana/create-plugin update` to get latest dev dependencies / configs / scripts etc
- Run `npm install` to migrate from yarn 1 to NPM
- Clean up dependencies to fix NPM peer dependency errors
- Delete `yarn.lock`

It also updates the github workflow:

- use NPM instead of yarn
- create a workflow matrix based on every plugin (plugin.json) to run builds, and canary builds, against all plugin examples
- conditionally run e2e tests if cypress directory exists in plugin example directory

Fixes: #167


mckn (approved):
LGTM! I did only test and try the app-basic but I guess all of them are migrated the same way.

Issue #169, opened 2023-05-29
Add unit tests to GH Actions


andresmgot:
Caveat: only two examples have unit tests. Running tests with no tests would result on an error.

PR #172, opened 2023-06-01
Unit tests

andresmgot:
Fixes https://github.com/grafana/grafana-plugin-examples/issues/169

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -1,5 +1,8 @@
 name: integration tests
 on:
+  push:
+    branches:
+      - "**"
   pull_request:
   schedule:
     - cron: "0 11 * * *" #once a day at 11 UTC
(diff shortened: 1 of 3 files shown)


andresmgot:
hey @jackw there is one example failing with the breaking change you noticed. Can we fix it or should we wait for a release?

jackw:
@andresmgot I think we need to update the app-with-extensions example [package.json](https://github.com/grafana/grafana-plugin-examples/blob/main/examples/app-with-extensions/package.json#L22) to use `"^10.1.0-118906pre"` for the `@grafana` dependencies otherwise it will continue to fail.

.github/workflows/integration-tests.yml
```diff
@@ -213,11 +226,24 @@ jobs:
...
         working-directory: ${{ matrix.pluginDir }}
 
+      - name: Test plugin backend
+        uses: magefile/mage-action@v2
+        with:
+          version: latest
+          args: -v test
+          workdir: ${{ matrix.pluginDir }}
+        if: steps.backend-check.outputs.MAGEFILE_EXISTS == 'true'
+
```
jackw, line 246: I'm not sure we need to run backend tests twice unless we're also bumping the go dependencies (which I don't think we're doing). 🤔

PR #175, opened 2023-06-20
Bumping docker compose files to make grafana_image flexible

tolzhabayev:
Related to https://github.com/grafana/plugin-tools/pull/286

Changes:
- Bumping `.config` folders via `create-plugin` `update` command
- Adjusting `docker-compose.yml` files to make docker image flexible

This is a non breaking change as it will always just fallback to default.


jackw (approved):
LGTM! 🚀

PR #179, opened 2023-07-18
Example: Add datasource-streaming-backend-websocket

### LINKED ISSUE #87: Examples: Rewrite example for how to implement web socket support in your plugin.
The current example for the datasource-streaming-websocket is implemented as a frontend only datasource plugin with an external backend. I don't think it is a useful example which is also highlighted in the following issue (https://github.com/grafana/grafana-plugin-examples/issues/2). 

There are also a couple of drawbacks with this approach when it comes to security since credentials to the external backend will be available in the browser.

What would have been better is to write an example where we show case how to push data from the backend part of the plugin to the frontend part. This example would be applicable for everyone trying to write a datasource plugin to any event based source.

oshirohugo:
Add a datasource plugin example capable of opening a stream connection to the backend and connect to a websocket service
Closes #87


examples/datasource-streaming-backend-websocket/streaming-backend-websocket-plugin/src/types.ts
```diff
@@ -0,0 +1,25 @@
...
+ * These are options configured for each DataSource instance
+ */
+export interface MyDataSourceOptions extends DataSourceJsonData {
+  uri?: string;
+}
+
+/**
+ * Value that is used in the backend, but never sent over HTTP to the frontend
+ */
+export interface MySecureJsonData {
```
andresmgot, line 23: this interface is not used

andresmgot (commented):
Great progress here! So far I have a few small suggestions.

I am no expert for the streaming feature but I tested the plugin and worked fine :+1:

examples/datasource-streaming-backend-websocket/streaming-backend-websocket-plugin/pkg/plugin/datasource.go
```diff
@@ -0,0 +1,191 @@
...
+
+	for {
+		select {
+		case <-ctx.Done():
+			return nil
+		case rawMsg := <-messages:
+			var msg Message
+			json.Unmarshal([]byte(rawMsg), &msg)
+
+			value := msg.Value*(d.upperLimit-d.lowerLimit) + d.lowerLimit
```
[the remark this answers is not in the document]
oshirohugo: `msg.Value` comes from the websocket connection. It's a random value, and the `upperLimit` and `loweLimit` are limit the range of the random value. My goal here was to exemplify how to pass data from the frontend to backend when a connection is made.

examples/datasource-streaming-backend-websocket/streaming-backend-websocket-plugin/pkg/plugin/datasource.go
```diff
@@ -0,0 +1,191 @@
...
+	var q Query
+	if err := json.Unmarshal(rawQuery, &q); err != nil {
+		return err
+	}
+
+	// simple validation of provided parameters
+	if q.UpperLimit > q.LowerLimit {
+		d.upperLimit = q.UpperLimit
+		d.lowerLimit = q.LowerLimit
+	}
```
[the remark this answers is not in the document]
oshirohugo: What would be the right way to pass a value from frontend to backend in a stream plugin after the plugin creation?

xnyo (commented):
Awesome work! :rocket: :rocket:   I tested it and it's working great.

I have left some comments about some typos and some missing error handling.

There's one issue with a goroutine that's not closed and I am not too sure how to solve it properly, just flagging it for now. I can take a better look next week :smiley:

examples/datasource-streaming-backend-websocket/streaming-backend-websocket-plugin/pkg/websocket/client.go
```diff
@@ -0,0 +1,65 @@
...
+			err := websocket.Message.Receive(c.conn, &message)
+			if err != nil {
+				Logger.Info("Error reading message: %v", err)
+				continue
+			}
+			messages <- message
+		}
+	}()
+	return messages
+}
```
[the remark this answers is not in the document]
oshirohugo: Nice catch! I solve this closing the connection from the caller when `ctx.Done()`. This will case an error on `websocket.Message.Receive(c.conn, &message)`, because the connection is close, then the loop will continue to the next case which is the `ctx.Done()`.
Here I changed a little bit your suggestion and `return` instead of `break` to avoid the `outer` label

examples/datasource-streaming-backend-websocket/streaming-backend-websocket-plugin/pkg/plugin/datasource.go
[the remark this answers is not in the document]
andresmgot: is it not part of the `RunStream` request? there is a `req *backend.RunStreamRequest`. This should be encoded somehow there.

Note that with this, two different queries for the same data source will edit the same data source properties (only the values of the last one executed will be saved).

examples/datasource-streaming-backend-websocket/streaming-backend-websocket-plugin/pkg/plugin/datasource.go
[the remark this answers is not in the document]
andresmgot: can you add a comment with this?

examples/datasource-streaming-backend-websocket/streaming-backend-websocket-plugin/pkg/websocket/client.go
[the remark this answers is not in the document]
oshirohugo: After some changes in the code, I've introduced one connection per query (like it was done in mqtt plugin) to fulfill Andres concerns about changing the datasource. 
Doing that I was having situation where the connection was close, to I would receive an err in `websocket.Message.Receive` but the `ctx.Done()` fired, so I've introduced a new channel `stopChan` that would be closed when the connection is closed and that would stop the loop

examples/datasource-streaming-backend-websocket/streaming-backend-websocket-plugin/src/datasource.ts
```diff
@@ -0,0 +1,14 @@
...
+import { MyQuery, MyDataSourceOptions, DEFAULT_QUERY } from './types';
+
+export class DataSource extends DataSourceWithBackend<MyQuery, MyDataSourceOptions> {
+  constructor(instanceSettings: DataSourceInstanceSettings<MyDataSourceOptions>) {
+    super(instanceSettings);
+  }
+
+  getDefaultQuery(app: CoreApp): Partial<MyQuery> {
+    return DEFAULT_QUERY;
+  }
```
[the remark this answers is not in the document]
oshirohugo: This is an easy way to have default values for my query, it's from `DataSourceApi`

examples/datasource-streaming-backend-websocket/streaming-backend-websocket-plugin/pkg/plugin/datasource.go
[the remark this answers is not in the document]
oshirohugo: I changed this, so each query will generate a different stream channel path and a different websocket connection. I'm taking the parameters from the channel path, like it's done in the MQTT plugin

examples/datasource-streaming-backend-websocket/streaming-backend-websocket-plugin/pkg/plugin/datasource.go
```diff
@@ -172,6 +167,32 @@ func (d *Datasource) PublishStream(context.Context, *backend.PublishStreamReques
...
 func (d *Datasource) RunStream(ctx context.Context, req *backend.RunStreamRequest, sender *backend.StreamSender) error {
+	chunks := strings.Split(req.Path, "/")
+	if len(chunks) < 3 {
+		return fmt.Errorf("invalid path: %s", req.Path)
+	}
+
+	q := Query{
+		LowerLimit: chunks[1],
+		UpperLimit: chunks[2],
+	}
```
andresmgot, line 178: this may work but I see that there ris a req.Data that seems like a better place to put query inputs. See for example how Loki is using it:

https://github.com/grafana/grafana/blob/main/pkg/tsdb/loki/streaming.go#L68

examples/datasource-streaming-backend-websocket/streaming-backend-websocket-plugin/src/components/QueryEditor.tsx
```diff
@@ -1,24 +1,33 @@
...
+    onRunQuery();
+  };
 
   const { upperLimit, lowerLimit } = query;
 
   return (
     <div className="gf-form">
       <InlineField label="Lower Limit" labelWidth={16} tooltip="Random numbers lower limit">
-        <Input onChange={handleEvent('lowerLimit', parseFloat)} onBlur={onRunQuery} value={lowerLimit || ''} />
+        <Input onChange={onLowerLimitChange} onBlur={onRunQuery} value={lowerLimit || ''} />
```
andresmgot, line 27: this should be `type="number"` btw

andresmgot (commented):
I tried this again and it seems that the inputs are being ignored? It always shows points between 0 and 1:

Screenshot from 2023-07-21 10-21-30

examples/datasource-streaming-backend-websocket/streaming-backend-websocket-plugin/pkg/plugin/datasource.go
oshirohugo: To be able to set `req.Data` I had to add frontend code. I've added a new implementation to `query` to create the stream. That made the `Query` implementation in the backend unnecessary to create the stream, so I removed it, which made the implementation of `backend.QueryDataHandler` unnecessary as well.

oshirohugo:
@andresmgot @xnyo this is ready for review again
Current changes:
- Added `query` implementation in `datasource.ts` (FE)
- Removed `Query` implementation in `datasource.go` (BE)
- Parsing query parameters from `req.Data` in `RunStream`
- Changed all query parameters both BE and FE to number instead of string

examples/datasource-streaming-backend-websocket/streaming-backend-websocket-plugin/src/datasource.ts
```diff
@@ -0,0 +1,42 @@
...
+
+import { defaults } from 'lodash';
+import { Observable, merge } from 'rxjs';
+
+import { MyQuery, MyDataSourceOptions, DEFAULT_QUERY } from './types';
+
+export class DataSource extends DataSourceWithBackend<MyQuery, MyDataSourceOptions> {
+  constructor(instanceSettings: DataSourceInstanceSettings<MyDataSourceOptions>) {
+    super(instanceSettings);
+  }
```
andresmgot, line 18: nit: if you are not modifying anything you can skip this?

andresmgot (commented):
Thanks for the changes! Code wise I only have some minor suggestions left.

In any case, it seems that the limits are still not working (at least for me). No matter what I introduce in the lower or upper limit, it still return values between 0-1. Am I missing something?

examples/datasource-streaming-backend-websocket/streaming-backend-websocket-plugin/src/datasource.ts
```diff
@@ -0,0 +1,42 @@
...
+
+  query(request: DataQueryRequest<MyQuery>): Observable<DataQueryResponse> {
+    const observables = request.targets.map((target, index) => {
+      const query = defaults(target, DEFAULT_QUERY);
+
+      return getGrafanaLiveSrv().getDataStream({
+        addr: {
+          scope: LiveChannelScope.DataSource,
+          namespace: this.uid,
+          path: `my-ws/custom-${query.lowerLimit}-${query.upperLimit}`, // this will allow each new query to create a new connection
```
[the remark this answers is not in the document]
oshirohugo: The path doesn't need to include the limit, but the path needs to change every time the query changes. A new query should create a new connection and the way of doing it is by creating a new path. A prettier way of doing the same thing is by creating a hash from the query parameters (that is what loki does, e.g.: 
https://github.com/grafana/grafana/blob/main/public/app/plugins/datasource/loki/streaming.ts#L68

https://github.com/grafana/grafana/blob/main/public/app/plugins/datasource/loki/streaming.ts#L21-L28)

I did like this because it seems simpler

examples/datasource-streaming-backend-websocket/streaming-backend-websocket-plugin/src/datasource.ts
oshirohugo: I've notice this code is comes from `npx @grafana/create-plugin@latest`.
Should we also remove it from the template later?

oshirohugo: Actually this is necessary indeed. I've removed it and the datasource frontend broke

oshirohugo:
> Thanks for the changes! Code wise I only have some minor suggestions left.
> 
> In any case, it seems that the limits are still not working (at least for me). No matter what I introduce in the lower or upper limit, it still return values between 0-1. Am I missing something?
Screenshot from 2023-07-24 14-25-35

It seems to be working, but I'm trying to reproduce your error

andresmgot (approved):
I rebuilt the images and now it's working as expected :+1: LGTM! Let's see if @xnyo finds something else.

examples/datasource-streaming-backend-websocket/streaming-backend-websocket-plugin/pkg/plugin/datasource.go
```diff
@@ -154,38 +93,65 @@ func (d *Datasource) CheckHealth(_ context.Context, req *backend.CheckHealthRequ
...
 func (d *Datasource) PublishStream(context.Context, *backend.PublishStreamRequest) (*backend.PublishStreamResponse, error) {
 	return &backend.PublishStreamResponse{
 		Status: backend.PublishStreamStatusPermissionDenied,
 	}, nil
 }
 
 func (d *Datasource) RunStream(ctx context.Context, req *backend.RunStreamRequest, sender *backend.StreamSender) error {
-	messages := d.wsClient.Read()
+	q := Query{}
+	json.Unmarshal(req.Data, &q)
```
xnyo, line 114: Missing error handling:

```suggestion
	if err := json.Unmarshal(req.Data, &q); err != nil {
		return fmt.Errorf("json unmarshal: %w", err)
	}
```

examples/datasource-streaming-backend-websocket/streaming-backend-websocket-plugin/pkg/plugin/datasource.go
```diff
@@ -154,38 +93,65 @@ func (d *Datasource) CheckHealth(_ context.Context, req *backend.CheckHealthRequ
...
 	for {
 		select {
 		case <-ctx.Done():
-			return nil
+			return ctx.Err()
 		case rawMsg := <-messages:
 			var msg Message
-			json.Unmarshal([]byte(rawMsg), &msg)
+			if err := json.Unmarshal([]byte(rawMsg), &msg); err != nil {
+				Logger.Error("Failed to unmarshal message", "error", err)
```
xnyo, line 141: Here it should skip execution of the body, so you should change this to:

```suggestion
				Logger.Error("Failed to unmarshal message", "error", err)
				break	
```

oshirohugo:
@andresmgot you were right, there was indeed a problem with the range. I could reproduce the error after changing the query multiple times. The connection was reused, which was causing many problems. Sometimes the connection would die out of nowhere and a re-connection would happen with default data, causing the issue you've mentioned. 
I changed the code, so every `RunStream` would create a new connection and the connection would be closed when `RunStream` returns

I've also simplified the error treatment. Since this is an example, for any error, I'm terminating the streaming. This should address the problem that @xnyo mentioned. But let me know if you prefer to add logs.

I've simplified the websocket client code and usage, by using `"github.com/gorilla/websocket"` instead of  `"golang.org/x/net/websocket"`. I was having many problem with it during my investigation  and according to its docs:
> This package currently lacks some features found in an alternative and more actively maintained WebSocket package

And gorilla seems to be back again!
https://github.com/gorilla/mux/issues/707#issuecomment-1629838706

andresmgot (approved):
:+1: cool, sounds good to me

examples/datasource-streaming-backend-websocket/streaming-backend-websocket-plugin/pkg/websocket/client.go
```diff
@@ -0,0 +1,44 @@
+package websocket
+
+import (
+	"github.com/gorilla/websocket"
+	"github.com/grafana/grafana-plugin-sdk-go/backend/log"
+)
+
+var Logger = log.DefaultLogger
```
xnyo, line 8: Nit: This variable is now unused

xnyo (approved):
Awesome work! LGTM!

PR #181, opened 2023-07-18
Add datasource streaming backend example

oshirohugo:
This PR adds a plugin example that has a stream connection to the backend.
This plugin is a minimal version, with only the necessary code to create a stream connection to the backend.
Compared to https://github.com/grafana/grafana-plugin-examples/pull/179 it has less noise as it doesn't add all the code necessary to an websocket connection.

## Note for reviwers:
This code is a subset of https://github.com/grafana/grafana-plugin-examples/pull/179 with some modifications:

- pkg/plugin/query.go: added one extra property `tickInterval` to control the frequency that datasource sends data to the backend
- pkg/plugin/datasource.go:
    * one more parameter is parsed in `RunStream`, and it's used to generated a random value which is sent as a frame value
    * a ticker is started to send frames with a given frequency
    * no websocket connection is opened and all the data is generate in this method
    * `CheckChannel` keeps the same code from the original template created by `create-plugin` tool
- src/datasource.ts: add tick interval to the path
- src/components/QueryEditor.tsx: add tick interval field
- src/components/ConfigEditor.tsx: all fields removed
- src/types.ts: updated to corresponding the added query fields and the removed configs


andresmgot:
can you highlight the code that has been simplified here? with so many files it's difficult to discover where the changes are

oshirohugo:
@andresmgot I've update the PR with the changes asked in https://github.com/grafana/grafana-plugin-examples/pull/179
and added a note for reviewer with the differences.
To save some time, let's just to the CR of this PR after the other one is done, because I maybe changes asked there could be applied here :D

oshirohugo:
@andresmgot I've update the PR with the last changes from https://github.com/grafana/grafana-plugin-examples/pull/179 and I also updated the highlighted differences in the "Note for reviewer"

andresmgot:
@oshirohugo this is a nice simplification :+1: I think having both example behaviors is nice (so people that doesn't need a websocket can focus on the simplified version). Having two different plugin examples may be confusing though for people looking for it since both examples are very similar (and would be more mainteinance work for us).

My suggestion is to combine this example with the one already merged, that way you can reuse code. For the UX, what I would do is, in the config editor, add a switch to use either the websocket connection or the mock data. I would move the `tickInterval` configuration to the ConfigEditor too so the QueryEditor (and datasource.ts code) is the same for both data-generation methods.

WDYT?

oshirohugo:
Since merging both examples would create more complexities to the code. I'm closing this PR

PR #185, opened 2023-07-31
Update all examples to 10.0.3 

academo:


scripts/update-grafana-version.sh
```diff
@@ -0,0 +1,107 @@
+#!/bin/bash
```
academo, line 1: The updates to package.json, docker and plugin.json files were made by this script.

scripts/update-grafana-version.sh
```diff
@@ -0,0 +1,107 @@
...
+                break
+            fi
+        done
+
+        if $upgradeFlag; then
+            upgradeCommand+="$dep@$GRAFANA_VERSION_TARGET "
+        else
+            upgradeCommand+="$dep@latest "
+        fi
+    done
```
xnyo, line 53: Nit: It looks like this block is indented twice

xnyo (approved):
LGTM! Awesome work! :rocket:

scripts/update-grafana-version.sh
academo: better safe than sorry!

PR #187, opened 2023-08-08
Add update-backend-sdk.sh script

xnyo:
Adds a script to update backend sdk version in all backend plugins.

It can be used standalone, but it is also added to `update-grafana-version.sh` at the end.

```bash
╰─❯ ./scripts/update-backend-sdk.sh --help
Usage: ./scripts/update-backend-sdk.sh [vX.Y.Z]
(defaults to latest version if not specified)
```

### DIFF
--- a/scripts/update-backend-sdk.sh
+++ b/scripts/update-backend-sdk.sh
@@ -8,14 +8,16 @@ fi
 
 SDK_TARGET="$1"
 
+# Keep track of base directory
+base=$(pwd)
+
 # Find go.mod files (backend plugins)
 files=$(find examples -type f -name "go.mod")
-
-base=$(pwd)
 
 # Iterate over each file and run `go get` to upgrade the grafana-plugin-sdk-go dependency
 for file in $files; do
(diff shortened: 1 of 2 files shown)


scripts/update-backend-sdk.sh
```diff
@@ -0,0 +1,27 @@
...
+base=$(pwd)
+
+# Find go.mod files (backend plugins)
+files=$(find examples -type f -name "go.mod")
+
+# Iterate over each file and run `go get` to upgrade the grafana-plugin-sdk-go dependency
+for file in $files; do
+  echo "Upgrading grafana-plugin-sdk-go to $SDK_TARGET in $file"
+
+  cd $(dirname "$file")
```
academo, line 21: bash trick. use `pushd $(dirname "$file") || exit 1` to enter a dir and `popd || exit 1` to go back to wherever  the shell was before the push, that way you don't need to keep track of the original CWD

academo (approved):

scripts/update-backend-sdk.sh
xnyo: Awesome, I applied the suggestion, thank you! :pray:

Issue #173, opened 2023-06-05
Use grafana/ui components in examples rather than referring to gf class names


marefr:
Use grafana/ui components in examples rather than referring to gf class names. Below an example that are using gf-classes and should be able to use [forms components](https://developers.grafana.com/ui/latest/index.html?path=/story/forms-form--async-validation):
https://github.com/grafana/grafana-plugin-examples/blob/a57c63cf9211df73b66bb0eb5f937b39262ffded/examples/datasource-http-backend/src/components/ConfigEditor.tsx#L14-L22

sympatheticmoose:
Couple of potential follow ups:
- validator check regex on this
- updates to docs to point towards correct components for styling

PR #190, opened 2023-08-11
plugin-examples:  

Ukochka:
**What this PR does**

- Uses grafana/ui components in examples rather than referring to gf class names in the components

- Deletes unused gf class names in the components

- swc/core version was changed to 1.3.75, because of the bug 'base_dir(./src) must be absolute. Please ensure that `jsc.baseUrl` is specified correctly.'

**Which task does it solve**
#173


academo (approved):
Great!

Issue #189, opened 2023-08-11
Switch to secureJsonFields for `app-with-backend` example (reference)


academo:
Introduced in https://github.com/grafana/grafana/pull/55313

Goal of this task is to switch to using `secureJsonFields` in https://github.com/grafana/grafana-plugin-examples/tree/main/examples/app-with-backend

You can reference to the PR linked for an example of what to change.

PR #201, opened 2023-09-13
Switch to secureJsonFields

oshirohugo:
### What it this PR about?
Use `secureJsonFields` in our examples instead extra property in `secureJson`
Solves https://github.com/grafana/grafana-plugin-examples/issues/189

### Note for reviwers:
https://github.com/grafana/grafana-plugin-examples/issues/189 asks for changes specifically in `app-with-backend`, but this plugin is not really using the `secureJsonFields` for anything.

The issue also asks to use the code suggested in https://github.com/grafana/grafana/pull/55313, but to allow the form to behave correctly the state property is necessary, therefore only the initial state value is changed to be taken from `secureJsonFields`. Make the form dependent only on a property won't allow the required form control.

### DIFF
--- a/examples/app-basic/provisioning/plugins/app.yaml
+++ b/examples/app-basic/provisioning/plugins/app.yaml
@@ -7,6 +7,5 @@ apps:
     disabled: false
     jsonData:
       apiUrl: http://default-url.com
-      isApiKeySet: true
     secureJsonData:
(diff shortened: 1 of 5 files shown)


academo (approved):

academo:
I wonder why we had that field instead of always check for a possible value.

PR #214, opened 2023-10-25
App-With-On-Behalf-Of: Inject `extended_jwt` conf in grafana ini file

gamab:
The `extended_jwt` config options are not overridable with env variables.
  
There were multiple ways of tackling this:
* changing Grafana to enable env overrides
* mounting a `custom.ini` file and using it instead of `/etc/grafana/grafana.ini`
* appending the `extended_jwt` config to the `/etc/grafana/grafana.ini`

I chose the last option so that it works with `grafana:main` already but also because we won't need to update the `custom.ini` file with grafana version increases.

Challenges are welcome.

### DIFF
--- a/examples/app-with-on-behalf-of-auth/.config/Dockerfile
+++ b/examples/app-with-on-behalf-of-auth/.config/Dockerfile
@@ -14,3 +14,7 @@ ENV GF_DEFAULT_APP_MODE "development"
 # Inject livereload script into grafana index.html
 USER root
 RUN sed -i 's/<\/body><\/html>/<script src=\"http:\/\/localhost:35729\/livereload.js\"><\/script><\/body><\/html>/g' /usr/share/grafana/public/views/index.html
+
+# Inject [extended_jwt] config in ini file
+COPY extended_jwt.ini /tmp/extended_jwt.ini
+RUN cat /tmp/extended_jwt.ini >> /etc/grafana/grafana.ini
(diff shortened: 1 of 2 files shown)


examples/app-with-on-behalf-of-auth/.config/Dockerfile
```diff
@@ -13,4 +13,8 @@ ENV GF_DEFAULT_APP_MODE "development"
 
 # Inject livereload script into grafana index.html
 USER root
-RUN sed -i 's/<\/body><\/html>/<script src=\"http:\/\/localhost:35729\/livereload.js\"><\/script><\/body><\/html>/g' /usr/share/grafana/public/views/index.html
\ No newline at end of file
+RUN sed -i 's/<\/body><\/html>/<script src=\"http:\/\/localhost:35729\/livereload.js\"><\/script><\/body><\/html>/g' /usr/share/grafana/public/views/index.html
+
+# Inject [extended_jwt] config in ini file
+COPY extended_jwt.ini /tmp/extended_jwt.ini
+RUN cat /tmp/extended_jwt.ini >> /etc/grafana/grafana.ini
```
andresmgot, line 20: I guess you could also copy the file as custom.ini? In any case both works so it's fine

andresmgot (approved):

andresmgot (approved):
:+1:

PR #223, opened 2023-11-23
Add filterQuery examples

andresmgot:
Add examples of how to filter out queries that are hidden or empty, which is something we recommend.


sunker (approved):
Nice!

PR #226, opened 2023-11-27
Use `iam` instead of `externalServiceRegistration`

gamab:
This PR aims at aligning example with changes introduced by https://github.com/grafana/grafana/pull/78686

PR #227, opened 2023-12-05
Update external service reg field to iam

wbrowne:
As per changes in https://github.com/grafana/grafana/pull/78686


andresmgot (commented):
Isn't this already there? https://github.com/grafana/grafana-plugin-examples/pull/226

PR #229, opened 2023-12-05
Edit of READMEs for app examples

josmperez:
Edit of several READMEs for Grafana style and readability. This PR will be followed by additional PRs for improving example docs.

### DIFF
--- a/examples/app-with-scenes/README.md
+++ b/examples/app-with-scenes/README.md
@@ -1,4 +1,4 @@
-# Grafana Scenes App plugin
+# Grafana Scenes App Plugin example
 
 This example demonstrates how to build a basic app plugin using @grafana/scenes framework.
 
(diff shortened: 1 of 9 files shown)


examples/app-with-scenes/README.md
```diff
@@ -1,14 +1,14 @@
-# Grafana Scenes App Plugin Template
+# Grafana Scenes App Plugin example
 
 This example demonstrates how to build a basic app plugin using @grafana/scenes framework.
 
 ## What are Grafana app plugins?
```
andresmgot, line 6: it may be worth it to review the original template too, for example: https://github.com/grafana/plugin-tools/blob/main/packages/create-plugin/templates/app/README.md?plain=1#L7 (which is using `datasources`, not  `data sources`).

andresmgot (approved):
LGTM

examples/app-with-scenes/README.md
josmperez: done.

PR #245, opened 2024-02-05
App Basic: Update react-router to v6

leventebalogh:
### What changed?

**Under app-basic:**
- Updated build configuration (using `create-plugin`)
- Updated to Grafana `10.3.1` (half manually)
- Updated to use react-router v6
- Removed the deprecated tab-bar based on `onNavChanged()`
- Updated the docs (in the README.md) accordingly 
- Updated one of the screenshots in the examples (README.md)

[View the updated README.md](https://github.com/grafana/grafana-plugin-examples/blob/leventebalogh/update-app-basic-to-react-router-v6/examples/app-basic/README.md)


examples/app-basic/README.md
```diff
@@ -81,15 +54,25 @@ export const MyComponent = () => {
 
 ---
 
-### Create a full-width page with no navigation bar
+### Create a full-width page
 
-You can hide the tab navigation bar and have access to the full view right to the Grafana sidebar by calling the `onNavChanged()` function with an `undefined` (as it is done in [utils.routing.ts#L25]).
+You can hide the tab navigation bar and have access to the full view right to the Grafana sidebar by changing the layout of the `<PluginPage>` component.
```
sunker, line 59: Maybe missing something, but this sentence confuses me. Maybe something like `You can hide the left sidebar and get access to the full width of the page by changing the layout of the `<PluginPage>` component.` would make it more clear?

examples/app-basic/src/components/App/App.tsx
```diff
@@ -1,13 +1,27 @@
...
+import { PageThree } from '../../pages/PageThree';
+import { PageFour } from '../../pages/PageFour';
+import { ROUTES } from '../../constants';
 
 export class App extends React.PureComponent<AppRootProps> {
   render() {
     return (
       <PluginPropsContext.Provider value={this.props}>
-        <Routes />
+        <Routes>
```
sunker, line 15: Rather than prefixing all routes' urls, can we provide a [basename](https://reactrouter.com/en/main/router-components/router) here instead?

sunker (approved):
Looks great, nice work! Just a nit and a question - nothing blocking.

examples/app-basic/README.md
leventebalogh: Yea, that sounds much better, will update it, thanks mate! 👍

examples/app-basic/src/components/App/App.tsx
leventebalogh: Hmm, as far as I understand the paths are relative by default in v6, so we don't need to prefix them or to set a basename - or did I misunderstand something?

examples/app-basic/README.md
leventebalogh: Fixed in https://github.com/grafana/grafana-plugin-examples/pull/245/commits/6acad4876b2fd4af0e5eb525b4b7ddb7eb27609e.

PR #246, opened 2024-02-08
Add example of an app with a WebAssembly dependency

sd2k:
Using WebAssembly in a plugin requires changing some webpack config. This adds an example of how to do so.

I've mainly made this as a reproducible example to demonstrate a bug in Grafana 10.2-10.4.0, but perhaps it's useful to have around as an e2e test.

Not sure how useful it really is but it was lying around as a draft, happy to close if you don't think it's needed!


andresmgot:
Adding @josmperez and @sympatheticmoose as a reviewers here to evaluate if we want this example. Sounds to me that it could be better documentation in the developer page.

academo:
if we decide to move on with this example we should make sure the integration tests are testing the real cases for this plugins.

We run these integrations tests daily to make sure the examples work and that the existing and upcoming changes in the grafana packages won't break plugins.

josmperez:
> if we decide to move on with this example we should make sure the integration tests are testing the real cases for this plugins.

In a recent conversation with David, he and I both expressed a desire in Q2 to better sync up and integrate our documentation, examples, and tutorials. As part of this process, we'll be streamlining our offerings and (I hope) making the examples we provide more interesting/useful.

With that in mind, is your comment a blocking issue right now, for this README PR, or can it wait until Q2?

josmperez:
> Adding @josmperez and @sympatheticmoose as a reviewers here to evaluate if we want this example. Sounds to me that it could be better documentation in the developer page.

In a recent conversation with David, he and I both expressed a desire in Q2 to better sync up and integrate our documentation, examples, and tutorials. Let's do some improvements/evaluation of this example in the near future.

examples/app-with-wasm/webpack.config.ts
```diff
@@ -0,0 +1,19 @@
...
+    // Add custom config here...
+    experiments: {
+      asyncWebAssembly: true,
+    },
+    output: {
+      publicPath: `public/plugins/${pluginJson.id}/`,
+    },
+  });
+};
+
```
jackw, line 18: I know this PR is quite old but it's worth mentioning that this isn't necessary any more with latest create-plugin configs it should work out the box.

sd2k: That's true! In fact I think enough things have been upstreamed into create-plugin that this PR is no longer really needed; everything should just work without any custom webpack config now :pray:

sd2k:
Closing because I don't think this is needed any more, the custom webpack stuff has been upstreamed into create-plugin's webpack config.

PR #248, opened 2024-02-21
AuthN: Remove app with on behalf of example

gamab:
We decided to drop the `externalServiceAuth` feature which embedded an OAuth2 server in Grafana to enable plugins to request OAuth2 tokens to query Grafana and to query on-behalf of users.

We provided a simpler alternative based on service account tokens  (see `externalServiceAccounts` feature). It does not cover the on-behalf of flow, this will eventually be covered with an `id_token` based solution.

This PR does the clean-up.


andresmgot (approved):
:+1:

PR #251, opened 2024-02-27
Chore: migrating app-basic to plugin-e2e

mckn:
### Description
This PR migrated the current cypress tests to plugin-e2e and playwright. I have stopped using test ids and relying on roles, texts etc. that are defined by the plugin instead.

Has a dependency on: https://github.com/grafana/plugin-tools/pull/795

### Todo
- [x] Rewrite tests.
- [x] Make the github action work with the plugin-e2e


examples/app-basic/playwright.config.ts
```diff
@@ -0,0 +1,56 @@
...
+    /* Collect trace when retrying the failed test. See https://playwright.dev/docs/trace-viewer */
+    trace: 'on-first-retry',
+  },
+
+  /* Configure projects for major browsers */
+  projects: [
+    {
+      name: 'chromium',
+      use: { ...devices['Desktop Chrome'] },
+    },
```
mckn, line 41: @sunker do I still need this one?

examples/app-basic/tests/appConfig.spec.ts
```diff
@@ -0,0 +1,23 @@
...
+  // reset the configured secret
+  await page.getByRole('button', { name: /reset/i }).click();
+
+  // enter some valid values
+  await page.getByLabel('API Key').fill('secret-api-key');
+  await page.getByLabel('API Url').clear();
+  await page.getByLabel('API Url').fill('http://www.my-awsome-grafana-app.com/api');
+
+  // listen for the server response on the saved form
+  const saveResponse = page.waitForResponse(`/api/plugins/${pluginJson.id}/settings`);
```
mckn, line 19: @sunker do you know of a better way to verify a successful saved config?

sunker (commented):
Looks like a great start! 

For App plugins, I think it would be good if the App itself defines models for all it's pages. Something along those lines: 

```ts
//AppConfigPage.ts
export class AppConfigPage extends GrafanaPage {
  constructor(ctx: PluginTestCtx) {
    super(ctx);
  }

  async save(options?: NavigateOptions) {
    const saveResponse = this.ctx.page.waitForResponse(this.ctx.selectors.pages.Apps.update(this.pluginId));
    await saveForm.click();
    return saveResponse;
  }
}

//appFixtures.ts
import { test as base } from '@grafana/plugin-e2e';

export const test = base.extend<{ appConfigPage: AppConfigPage }>({
  appConfigPage: ({ page }, use) => {
    await use(new AppConfigPage({ page, selectors, grafanaVersion, request });
  },
});

//appConfigPage.spec.ts
import { test } from './appFixtures'
test('configuring app', async ({ appConfigPage }) => {
 ...
})
```

Possibly, plugin-e2e could provide an BaseAppConfigPage model that provide some basic methods such as `Save`? Then app plugins could extend it and add its own stuff.

examples/app-basic/tests/appConfig.spec.ts
mckn: Referencing this PR: https://github.com/grafana/plugin-tools/pull/795
Will update this test when that one is available.

.github/workflows/integration-tests.yml
```diff
@@ -229,7 +240,7 @@ jobs:
 
       - name: Upgrade @grafana packages using canary_version
         run: |
-          npm install --legacy-peer-deps $(echo $(cat package.json | jq -r --arg version $CANARY_VERSION '["@grafana/eslint-config", "@grafana/tsconfig", "@grafana/scenes", "@grafana/experimental"] as $blacklist | [(.devDependencies,.dependencies) | keys] | flatten | unique | map(select ( test("^@grafana") ) ) as $deps | $deps - $blacklist | map("\(.)@\($version)") | join(" ")') )
+          npm install --legacy-peer-deps $(echo $(cat package.json | jq -r --arg version $CANARY_VERSION '["@grafana/eslint-config", "@grafana/tsconfig", "@grafana/scenes", "@grafana/experimental", "@grafana/plugin-e2e"] as $blacklist | [(.devDependencies,.dependencies) | keys] | flatten | unique | map(select ( test("^@grafana") ) ) as $deps | $deps - $blacklist | map("\(.)@\($version)") | join(" ")') )
           npm install --force --save-optional @swc/core  @swc/core-linux-arm-gnueabihf @swc/core-linux-arm64-gnu @swc/core-linux-arm64-musl @swc/core-linux-x64-gnu @swc/core-linux-x64-musl
         working-directory: ${{ matrix.pluginDir }}
```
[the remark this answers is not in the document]
mckn: Yeah, so I started to look into that but I'm not sure we want to use it here. Or at least we need to have a discussion about it in the squad.

So currently the integration tests run daily to verify that the examples are still working against the latest version of Grafana (which has changed). So with that setup it does not make any sense to re-run for older version every day.
But it might make sense to change this so we run it when a version changed regardless if it is the latest version or not and base it on the `grafanaDependency` flag for each example instead.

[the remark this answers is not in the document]
mckn: Short answer is Yes! Agree - but I think we need to redesign this workflow a bit before we can do that so I will probably not cover that in this PR.

examples/app-basic/tests/appNavigation.spec.ts
```diff
@@ -0,0 +1,31 @@
+import pluginJson from '../src/plugin.json';
+import { test, expect } from '@grafana/plugin-e2e';
+
+test.describe('navigating app', () => {
```
mckn, line 4: I was thinking about wrapping this in some kind of page model as well but feels a bit overkill, wdyt @sunker?

examples/app-basic/tests/fixtures.ts
```diff
@@ -0,0 +1,19 @@
+import { AppConfigPage, test as base } from '@grafana/plugin-e2e';
+import pluginJson from '../src/plugin.json';
+
+type AppTestFixture = { appConfigPage: AppConfigPage };
+
+export const test = base.extend<AppTestFixture>({
```
mckn, line 6: I like that we are able to define the pluginId once and then can reuse it for all tests.

sunker: Yes this is really neat 👍

examples/app-basic/tests/appConfig.spec.ts
```diff
@@ -0,0 +1,22 @@
+import { test, expect } from './fixtures';
+
+test('configuring app', async ({ appConfigPage, page }) => {
+  await appConfigPage.goto();
```
sunker, line 4: Can we move the `goto` method call to the fixture (we do that for all other page fixtures)?

examples/app-basic/tests/appConfig.spec.ts
```diff
@@ -0,0 +1,22 @@
...
+
+  const saveForm = page.getByRole('button', { name: /Save API settings/i });
+  await expect(saveForm).toBeVisible();
```
sunker, line 7: Unlike in cypress, you generally don't need to check visibility. All Playwright actions such as the click method below performs [actionability](https://playwright.dev/docs/actionability#introduction) checks that will do retries until it succeeds or fails.

examples/app-basic/tests/appConfig.spec.ts
```diff
@@ -0,0 +1,22 @@
...
+
+  // reset the configured secret
+  await page.getByRole('button', { name: /reset/i }).click();
+
+  // enter some valid values
+  await page.getByLabel('API Key').fill('secret-api-key');
+  await page.getByLabel('API Url').clear();
+  await page.getByLabel('API Url').fill('http://www.my-awsome-grafana-app.com/api');
```
sunker, line 15: Nit, but would be good to interact with these form elements using the approach that we suggest in the [docs](https://grafana.com/developers/plugin-tools/e2e-test-a-plugin/selecting-elements#input-field). Should be `getByRole` for inputs.

examples/app-basic/tests/appNavigation.spec.ts
sunker: Yeah probably overkill if we only want to verify that they render successfully.

sunker (commented):
Looks great! Just some minor comments.

.gitignore
```diff
@@ -3,4 +3,13 @@ node_modules
...
+.DS_Store
+**/test-results/
+**/playwright-report/
+**/blob-report/
+**/playwright/.cache/
+**/test-results/
+**/playwright-report/
+**/blob-report/
+**/playwright/.cache/
+**/playwright/.auth/
```
jackw, line 15: I'm possibly missing something but I can't see anything added in this PR that lives in any of these directories plus there seems to be a bit of repetition. Maybe we can trim this down to the following?

```suggestion
**/blob-report/
**/playwright-report/
**/playwright/
**/test-results/
```

examples/app-basic/.gitignore
```diff
@@ -29,4 +29,13 @@ e2e-results/
...
+.eslintcache
+/test-results/
+/playwright-report/
+/blob-report/
+/playwright/.cache/
+/test-results/
+/playwright-report/
+/blob-report/
+/playwright/.cache/
+/playwright/.auth/
```
jackw, line 41: Same suggestion as root gitignore:

```suggestion
/blob-report/
/playwright-report/
/playwright/
/test-results/
```

examples/app-basic/package.json
```diff
@@ -5,8 +5,7 @@
   "scripts": {
     "build": "webpack -c ./.config/webpack/webpack.config.ts --env production",
     "dev": "webpack -w -c ./.config/webpack/webpack.config.ts --env development",
-    "e2e:update": "npm exec cypress install && npm exec grafana-e2e run --update-screenshots",
-    "e2e": "npm exec cypress install && npm exec grafana-e2e run",
+    "e2e": "npx playwright test",
```
jackw, line 8: I don't think we need to use npx here. Additionally something we need to consider is if anyone uses this as a "starter" for building an app create-plugin updates (as of now) will replace/remove this script.

```suggestion
    "e2e": "playwright test",
```

examples/app-basic/src/components/App/App.tsx
```diff
@@ -5,19 +5,17 @@ import { PluginPropsContext } from '../../utils/utils.plugin';
...
     return (
       <PluginPropsContext.Provider value={this.props}>
         <Routes>
-          <Route path={ROUTES.Two} element={<PageTwo />} />
-          <Route path={`${ROUTES.Three}/:id?`} element={<PageThree />} />
+          <Route path={`${ROUTES.Two}/:id?`} element={<PageTwo />} />
 
           {/* Full-width page (this page will have no navigation bar) */}
-          <Route path={ROUTES.Four} element={<PageFour />} />
+          <Route path={ROUTES.Three} element={<PageThree />} />
```
jackw, line 18: What happened to page 4?

jackw (commented):
Lovely stuff removing all the excess index.ts files for routes and testids! 👏 

Left a few nits. Does this PR mean plugin-e2e is almost ready for a 1.0.0 release?

examples/app-basic/src/components/App/App.tsx
mckn: I just killed it :)

mckn: Didn't feel like it brought any value since we have changed how the navigation works.

sunker (approved):
Awesome work on this! 

Would be great if you could apply [this](https://github.com/grafana/grafana-plugin-examples/pull/251#discussion_r1531598425) comment so this page follow the same convention as all the other fixtures.

mckn:
> Awesome work on this!
> 
> Would be great if you could apply [this](https://github.com/grafana/grafana-plugin-examples/pull/251#discussion_r1531598425) comment so this page follow the same convention as all the other fixtures.

Sorry, missed that one - fixed!

PR #252, opened 2024-03-14
Chore: migrating app-basic-with-backend to plugin-e2e

mckn:
### Description
This PR migrated the current cypress tests to plugin-e2e and playwright. I have stopped using test ids and relying on roles, texts etc. that are defined by the plugin instead.

Has a dependency on: https://github.com/grafana/plugin-tools/pull/821

### Todo
- [x] Rewrite tests.


examples/app-with-backend/tests/appCallingBackend.spec.ts
```diff
@@ -0,0 +1,17 @@
+import { test, expect } from "./fixtures";
+
+test("page one should successfully call backend and display reply", async ({
+  pageOne,
+  page,
+}) => {
+  await pageOne.goto();
```
sunker, line 7: Would be good to move the `goto` method call to the fixture (that's what we do in all other page fixtures)

sunker (approved):
Looks great! Just a nit.

PR #255, opened 2024-03-20
Update query filtering examples 

sunker:
Now that [query filtering](https://github.com/grafana/grafana/pull/84656) is being changed in Grafana, it's time to update our examples. There's no longer a need to filter out hidden queries in the data source as this is being taken care of by Grafana. Also, `filterQuery` is now being called from the query runner, so frontend data sources no longer need to do this in the query method. 

While at it, I'm also replacing usage of `defaults(target, defaultQuery)` with the implementation of `getDefaultQuery`.

### DIFF
--- a/examples/datasource-streaming-websocket/streaming-websocket-plugin/src/DataSource.ts
+++ b/examples/datasource-streaming-websocket/streaming-websocket-plugin/src/DataSource.ts
@@ -1,5 +1,6 @@
 import {
   CircularDataFrame,
+  CoreApp,
   DataQueryRequest,
   DataQueryResponse,
   DataSourceApi,
@@ -7,7 +8,6 @@ import {
   FieldType,
   LoadingState,
 } from '@grafana/data';
-import defaults from 'lodash/defaults';
 import { merge, Observable } from 'rxjs';
 import { defaultQuery, MyDataSourceOptions, MyQuery } from './types';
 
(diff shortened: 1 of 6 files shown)


examples/app-basic/yarn.lock
andresmgot, line 1: why are there updates in the `yarn.lock`s?

examples/datasource-streaming-websocket/streaming-websocket-plugin/src/DataSource.ts
```diff
@@ -65,9 +67,6 @@ export class DataSource extends DataSourceApi<MyQuery, MyDataSourceOptions> {
   }
 
   filterQuery(query: MyQuery): boolean {
-    if (query.hide) {
-      return false;
-    }
-    return true;
+    return !!query.queryText;
```
andresmgot, line 70: I think queryText is not used in this plugin

yarn.lock
```diff
@@ -0,0 +1,4 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
andresmgot, line 1: this file is probably not needed

andresmgot (approved):
LGTM

examples/datasource-streaming-websocket/streaming-websocket-plugin/src/DataSource.ts
sunker: Maybe it shouldn't be, but it is [defined](https://github.com/grafana/grafana-plugin-examples/blob/375cc7ec9eae4f7d712172b13a361ef6818d0377/examples/datasource-streaming-websocket/streaming-websocket-plugin/src/types.ts#L4) in types.

PR #256, opened 2024-03-21
Remove usage of gf-forms

sunker:


andresmgot (approved):

examples/datasource-logs/src/components/QueryEditor.tsx
```diff
@@ -12,18 +12,17 @@ export function QueryEditor({ query, onChange }: Props) {
   };
 
   const onQueryLimitChange = (event: ChangeEvent<HTMLInputElement>) => {
-    onChange({ ...query, limit: parseInt(event.target.value,10) });
+    onChange({ ...query, limit: parseInt(event.target.value, 10) });
   };
 
   return (
-    <div className="gf-form">
+    <HorizontalGroup spacing="none">
```
jackw, line 19: I've a feeling we should be using `[<Stack />](https://developers.grafana.com/ui/latest/index.html?path=/story/general-layout-stack--direction-examples)` these days rather than these older layout components.

jackw (approved):
Nice one. Left a nit as I think HorizontalLayout is being phased out in favour of stack.

examples/datasource-logs/src/components/QueryEditor.tsx
sunker: Good shout 👍

PR #260, opened 2024-04-02
Panel datalinks: Add plugin e2e tests and remove cypress 

sunker:
This PR adds e2e tests using the @grafana/plugin-e2e package. Also removing cypress tests and its dependencies.


examples/panel-datalinks/tests/panel.spec.ts
```diff
@@ -0,0 +1,46 @@
...
+import { testIds } from '../src/components/testIds';
+
+test.describe('panel-datalinks panel', () => {
+  let panelEditPage: PanelEditPage;
+  test.beforeEach(async ({ gotoPanelEditPage }) => {
+    panelEditPage = await gotoPanelEditPage({ dashboard: { uid: 'fTh-POZ4k' }, id: '2' });
+  });
+
+  test('should display "No data" when no data is passed to the panel', async ({ page }) => {
+    await expect(page.getByTestId(testIds.panel.noData)).not.toBeVisible();
```
mckn, line 11: Q: should we base the tests on what the user sees instead of the test id as we did before? So maybe check for the No data label instead?

mckn (approved):
LGTM! One small nit regarding the selector(s).

examples/panel-datalinks/tests/panel.spec.ts
sunker: yes good shout! will fix

PR #264, opened 2024-04-08
DataSource Basic: Move to Playwright and plugin-e2e

sunker:
This pr does the following:
* Removes cypress tests
* Removes grafana/e2e deps
* Adds config editor and query editor tests
* Remove legacy form fields
* Align health check, components and types with what's being scaffolded with create-plugin


examples/datasource-basic/playwright.config.ts
```diff
@@ -0,0 +1,52 @@
...
+      testMatch: [/.*\.js/],
+    },
+    // 2. Run tests in Google Chrome. Every test will start authenticated as admin user.
+    {
+      name: 'chromium',
+      use: { ...devices['Desktop Chrome'], storageState: 'playwright/.auth/admin.json' },
+      dependencies: ['auth'],
+    },
+  ],
+});
```
leventebalogh, line 52: Would it make sense to have this basic config under the `.config/` folder as well, and only extend that one in this file?

(Like this we could make it easier in the future to make changes to the base config, and users could also extend it without interfering with our future changes.)

examples/datasource-basic/src/components/ConfigEditor/ConfigEditor.tsx
```diff
@@ -1,46 +1,48 @@
...
+      <InlineField
+        label="Time Field"
+        labelWidth={14}
+        interactive
+        tooltip="Default time field used when interpolate the $__timeFilter()"
+      >
+        <Input onChange={onTimeFieldChange} placeholder="time" value={jsonData?.defaultTimeField ?? ''} width={40} />
+      </InlineField>
+      <InlineField label="API Key" labelWidth={14} interactive tooltip={'Secure json field (backend only)'}>
+        <SecretInput
```
leventebalogh, line 35: 👏

leventebalogh (approved):
LGTM 👍

examples/datasource-basic/playwright.config.ts
leventebalogh: Hmm, yeah, that's probably a good point. (Something that we should probably discuss later is - should we provide a basic playwright config under .config/ for new scaffolds, or just leave playwright out from this pattern?)

sunker: Even though not ideal to have a different pattern for Playwright, I think it's worth making an exception here and leave the playwright conf in the root. If not, it can become pretty messy if for example we'd like the `update` cmd to make changes to the playwright config. 

Any thoughts @jackw?

leventebalogh: Yeah, I am not against making an exception in this case. (How do we imagine the `update` cmd updating the playwright config in case it sits in the plugin root directory?)

sunker: I'm not entirely sure tbh. Actually I'm struggling to see when we would ever want to update the Playwright config. The plugin author should have complete control over the config.

PR #269, opened 2024-04-09
App with extensions: Move to Playwright and plugin-e2e

sunker:
This pr does the following:
* Removes cypress tests
* Removes grafana/e2e deps
* Migrates existing Cypress tests to use Playwright and plugi-e2e


leventebalogh (approved):
Nice 👏

PR #270, opened 2024-04-09
App with extensions point: Move to Playwright and plugin-e2e

sunker:
This pr does the following:
* Removes cypress tests
* Removes grafana/e2e deps
* Migrates existing Cypress tests to use Playwright and plugi-e2e


examples/app-with-extension-point/package.json
```diff
@@ -5,8 +5,7 @@
   "scripts": {
     "build": "webpack -c ./.config/webpack/webpack.config.ts --env production",
     "dev": "webpack -w -c ./.config/webpack/webpack.config.ts --env development",
-    "e2e:update": "npm exec cypress install && npm exec grafana-e2e run --update-screenshots",
-    "e2e": "npm exec cypress install && npm exec grafana-e2e run",
+    "e2e": "npx playwright test",
```
mckn, line 8:
```suggestion
    "e2e": "playwright test",
```

I don't think we need to run it via npx

examples/app-with-extension-point/tests/extendActions.spec.ts
```diff
@@ -0,0 +1,18 @@
+import { test, expect } from '@grafana/plugin-e2e';
```
mckn, line 1: Was thinking about wether or not to keep the testIds. What do you think about replacing them with other selectors such as getByRole or getByLabel?

mckn (commented):
LGTM! Added a couple of questions.

examples/app-with-extension-point/tests/extendActions.spec.ts
sunker: There are no form elements where labels are associated with inputs. Could potentially use getByRole in some places, but I don't see any harm in using test ids here.

leventebalogh (approved):
LGTM 👍

PR #272, opened 2024-04-11
Remove cypress entirely 

sunker:
Now that all example plugins that had cypress tests have been migrated to use Playwright, it's time to remove Cypress entirely from this repo. What this PR does:
* remove cypress smoke tests 
* remove cypress related files and directories from gitignore 
* removes grafana/e2e script

Most importantly, this PR updates the `integration-tests.yml` workflow. All cypress references are now gone and I have added a few extra steps such as uploading the Playwright report to GCS. Note that the Playwright report will be public on the Internet, but I think it should be fine as we don't depend on secrets in the e2e tests for the example plugins in this repo.


examples/datasource-http-backend/package.json
```diff
@@ -19,8 +17,6 @@
   "license": "Apache-2.0",
   "devDependencies": {
     "@babel/core": "^7.21.4",
-    "@grafana/e2e": "10.0.3",
-    "@grafana/e2e-selectors": "10.0.3",
```
leventebalogh, line 23: **Just a question:** should we keep these scripts but using playwright now, so they match with what we have [in our templates](https://github.com/grafana/plugin-tools/blob/main/packages/create-plugin/templates/common/_package.json#L13)? I guess probably not super important, as we don't have any meaningful tests, only in a case if someone would check out this folder and start out a plugin from it.

leventebalogh (approved):
LGTM 👍

examples/datasource-http-backend/package.json
sunker: Do you mean the `e2e` script? Probably fine to remove I guess - the [migration guide](https://grafana.com/developers/plugin-tools/e2e-test-a-plugin/migrate-from-grafana-e2e#3-update-scripts) has instructions on what scripts to add.

PR #277, opened 2024-04-19
Update grafana packages to 10.3.1

### LINKED ISSUE #276: Update grafana/ui to 10.3.1 for all plugins examples
To be able to use new Layout components like Stack we need to update grafana/ui to 10.2.3 version

https://raintank-corp.slack.com/archives/C024NL6AEJH/p1713355923681009?thread_ts=1713342336.732319&cid=C024NL6AEJH

Ukochka:
To be able to use new grafana/ui we need to update to 10.2.3


examples/app-basic/package.json
```diff
@@ -57,10 +57,10 @@
   },
   "dependencies": {
     "@emotion/css": "^11.1.3",
-    "@grafana/data": "10.3.1",
-    "@grafana/runtime": "10.3.1",
-    "@grafana/schema": "10.3.1",
-    "@grafana/ui": "10.3.1",
+    "@grafana/data": "10.2.3",
```
jackw, line 60: This is downgrading dependencies in certain examples. I don't think we want to do that. 🤔

examples/app-with-rbac/src/plugin.json
```diff
@@ -52,21 +60,33 @@
...
       },
-      "grants": ["Viewer"]
+      "grants": [
+        "Viewer"
+      ]
     }
   ],
   "dependencies": {
-    "grafanaDependency": ">=10.4.0",
+    "grafanaDependency": ">=10.3.1",
```
academo, line 89: Is this downgrade intended?

academo:
@Ukochka apart from the pointed downgrade we are missing that the docker-compose file use the correct grafana version. We are specifying now that the grafana version is 10.3.x but most of them use a lower version of grafana.

Ukochka:
@academo thank you for pointing out! Fixed!

academo (approved):

PR #279, opened 2024-04-24
Examples: cleaning up tests for app basic a bit more

mckn:
## Description

Cleaned up the navigation tests a bit more. By introducing a `gotoPage` function we can remove the need of suppling plugin id in each test. This is only done once in the fixture.

### DIFF
--- a/examples/app-basic/tests/fixtures.ts
+++ b/examples/app-basic/tests/fixtures.ts
@@ -13,13 +13,14 @@ export const test = base.extend<AppTestFixture>({
     });
     await use(configPage);
   },
-  gotoPage: async ({ gotoAppPage }, use) =>
-    use((path) =>
+  gotoPage: async ({ gotoAppPage }, use) => {
+    await use((path) =>
       gotoAppPage({
         path,
         pluginId: pluginJson.id,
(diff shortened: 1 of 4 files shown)


examples/app-basic/tests/fixtures.ts
```diff
@@ -14,6 +17,13 @@ export const test = base.extend<AppTestFixture>({
     await configPage.goto();
     await use(configPage);
   },
+  gotoPage: ({ gotoAppPage }, use) =>
+    use((path) =>
+      gotoAppPage({
```
sunker, line 22: Need to return our await this invocation

sunker (approved):
Looks like a nice improvement! Just a nit. 

Please update the templates in create-plugin too. :)

examples/app-basic/tests/fixtures.ts
mckn: It does return since I don't have any `{}` around the `gotoPage` function. But I can wrap it and return if it makes it more explicit.

mckn:
> Please update the templates in create-plugin too. :)

On it! :)

PR #280, opened 2024-04-25
Update plugin-e2e dep in ci

sunker:
Updates the plugin-e2e package so that we always run e2e tests using the latest minor/patch in CI.

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -74,6 +74,11 @@ jobs:
           npm ci
         working-directory: ${{ matrix.pluginDir }}
 
+      - name: Update plugin-e2e
+        run: |
+          npm update @grafana/plugin-e2e
+        working-directory: ${{ matrix.pluginDir }}
+
       - name: Run frontend tests
         run: |
           npm run test:ci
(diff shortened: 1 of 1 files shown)


tolzhabayev (approved):
it works, but probably we should think about solving this in examples themselves additionally.

sunker:
> it works, but probably we should think about solving this in examples themselves additionally.

How could we do that?

tolzhabayev:
> > it works, but probably we should think about solving this in examples themselves additionally.
> 
> How could we do that?

dependabot could probably do that - open PRs with updates to this package.

jackw:
> dependabot could probably do that - open PRs with updates to this package. We can theoretically even auto merge those specific ones.

I think this option gets my vote too. Whilst the changes in this PR work it gives us the idea that our e2e tests pass fine but when we run them locally or a developer copies the example as a starting point they'll fail which might cause confusion.

sunker:
> dependabot could probably do that - open PRs with updates to this package.

Yes this makes sense to me. I'll have a look at this

sunker:
Closing this PR in favour of https://github.com/grafana/grafana-plugin-examples/pull/282. Please take a look when you have time @tolzhabayev @jackw.

PR #281, opened 2024-04-25
AppWithDashboard: updating tests to use navigation function

mckn:
## Description
This PR will update the `app-with-dashboard` tests to create a fixture that will define the `gotoPage` navigation function. The benefit of this is that you don't need to repeat the `/a/${pluginId}` in every test.


leventebalogh (approved):
LGTM 👍

PR #282, opened 2024-04-26
Update dependabot conf

sunker:
This PR changes the dependabot conf in the following way:
* correcting the directory path to each example
* adds update configs for those plugin examples that weren't previously included
* for each example, adding yet another update config that checks for plugin-e2e updates more frequently (daily as opposed to monthly) 

Question: I guess since the path to each example was not right, we're going to see more chore version updates. These are limited to 5 PR at the time, which means the grafana deps can be out of sync. Can we define a group for the grafana deps (exc plugin-e2e)? Or since we have the grafana version update [script](https://github.com/grafana/grafana-plugin-examples/blob/main/scripts/update-grafana-version.sh), should we simply ignore the grafana deps in the dependabot config? 

I'll look into adding an auto-merging workflow once I've seen this running successfully in main for some time.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -9,7 +9,7 @@ updates:
       - "grafana/plugins-platform"
 
   - package-ecosystem: "npm"
-    directory: "/app-basic"
+    directory: "/examples/app-basic"
     schedule:
       interval: "monthly"
     open-pull-requests-limit: 5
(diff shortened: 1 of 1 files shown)


sunker:
As discussed in the daily, I'm going to look into replacing depandabot with renovate as it's more flexible and capable of handling group updates (for grafana ui, data and runtime) and it supports globs (updating all example plugin that uses plugin-e2e in one single pr). 

Closing this pr.

PR #288, opened 2024-05-08
Update all the plugins using create-plugin update

Ukochka:
History:
There was a script which @academo wrote to update grafana packages. It took desired grafana version and updated specific packages to this version.

Work:
While fixing problems for mac OS I added some improvements to check first if the current version is newer than target version and leave it as it is to avoid downgrading

Latest updates:
But we decide to use `npx create-plugin@latest update` to update plugins from now on, because in the future we want to make it automatically. So this PR is a first step to do it.
Using create-plugin update tool we update not only npm packages but also config files


jackw:
I hate to ask this but given the size of this PR is it possible we could separate it out into individual plugin example update PRs to make it easier for reviewers?

Ukochka:
This PR will be splited to many smaller PRs

PR #304, opened 2024-05-23
Delete unused e2e dependencies

Ukochka:
After create-plugin update it was e2e added but it is not used in any of plugins examples


sunker (approved):
Nice

PR #309, opened 2024-05-27
Delete panel plugins: flot, plotly, scatterplot, visx

Ukochka:
After [suggestion](panel-flot
panel-plotly
panel-scatterplot
panel-visx) from @jackw we decided that we do not need panel-flot, panel-plotly, panel-scatterplot, panel-visx


leventebalogh (approved):
LGTM 👍

PR #312, opened 2024-05-30
Update README.md

eleijonmarck:


examples/app-with-rbac/README.md
```diff
@@ -13,7 +13,7 @@ This App plugin example shows you how to leverage Grafana RBAC to control access
 Your development environment must meet the following prerequisite:
 
 - Grafana version 10.4.0 or later.
-- You must enable RBAC by setting the `accessControlOnCall` feature toggle.
+- You must enable the following features by setting the `accessControlOnCall,idForwarding,externalServiceAccounts` feature toggle.
```
gamab, line 16: That's not exactly accurate, you can use accessControlOnCall only. idForwarding and externalServiceAccounts are only needed if you have a plugin backend as written further down.

PR #315, opened 2024-06-05
Playwright tests for datasource plugins examples

Ukochka:
Tests for:
datasource-logs
datasource-streaming-websocket
datasource-streaming-backend-websocket


examples/datasource-streaming-websocket/streaming-websocket-plugin/provisioning/datasources/datasources.yml
```diff
@@ -0,0 +1,15 @@
+apiVersion: 1
```
academo, line 1: are you missing the URL in this provisioning file?

It is configured manually in the e2e tests but if someone were to use this provisioning to test the example the datasource won't work.

examples/datasource-logs/src/components/QueryEditor.tsx
```diff
@@ -18,10 +18,16 @@ export function QueryEditor({ query, onChange }: Props) {
   return (
     <Stack gap={0}>
       <InlineField label="Query" labelWidth={20} tooltip="Query text">
-        <Input onChange={onQueryTextChange} value={query.queryText || ''} width={70} />
+        <Input id="query-editor-query-text" onChange={onQueryTextChange} value={query.queryText || ''} width={70} />
```
academo, line 21: I see you are using `test-id` in the other query editor tests but `id` here. Can we keep it consistent? I think the `test-id` is the best way to go

academo (approved):

PR #316, opened 2024-06-07
Add e2e tests for panel-frame-select

Ukochka:
I added `<PanelDataErrorView` to SimplePanel to display "No data" message as at all other plugins
I added 2 tests to check if there is no data and if there is data


sunker (approved):
LGTM

You need to add the `e2e` script in package.json to get the tests to work in CI. See [this](https://github.com/grafana/grafana-plugin-examples/blob/cf802d7befc642abe06493a3688a70798b75fe76/examples/panel-basic/package.json#L8) example.

PR #318, opened 2024-06-13
e2e tests for http-backend-datasource

Ukochka:
To be able to run e2e tests at CI we need mockserver which will return values for our tests. 

I updated grafana packages to 10.4.0 because change was introduced here https://github.com/grafana/grafana/pull/79615 DataSourceHttpSettings  component is using e2e selector instead of aria-label

It also fixes the problem of running e2e tests on current and latest grafana versions

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -151,7 +151,7 @@ jobs:
           fi
         working-directory: ${{ matrix.pluginDir }}
 
-      - name: Run mock server
+      - name: Run mockserver
         if: steps.mockserver-check.outputs.MOCKSERVER_DEFINED == 'true'
         run: docker compose up -d mockserver
         working-directory: ${{ matrix.pluginDir }}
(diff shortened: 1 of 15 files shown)


.github/workflows/integration-tests.yml
```diff
@@ -139,6 +139,23 @@ jobs:
           npm exec --package semver-compare-cli --call 'semver-compare $LATEST_STABLE_VERSION ge $EXPECTED_GRAFANA_VERSION'
         working-directory: ${{ matrix.pluginDir }}
 
+      - name: Check if mockserver is defined
+        id: mockserver-check
+        run: |
+          echo "hello?"
```
sunker, line 145: hello :)

sunker (approved):
Great stuff!

.github/workflows/integration-tests.yml
Ukochka: Will delete that :)

PR #319, opened 2024-06-13
Datasource-http-backend: Delete unused server code and update documentation

Ukochka:
Server was rewritten and this code is not valid anymore


examples/datasource-http-backend/README.md
```diff
@@ -45,14 +43,20 @@ $ mage -v
...
 ### Example server
 
+mockserver is included to docker-compose file. You can run or run server separetaly 
+
 ```bash
-$ mage server
-$ ./cmd/server/server :10000
-2022/10/28 15:43:16 listening on :10000
+cd server
+docker-compose up
```
academo, line 50: Since this section talks only about the server this command should be `docker compose up mockserver`

examples/datasource-http-backend/README.md
```diff
@@ -45,14 +43,20 @@ $ mage -v
...
 ```
 
 Then, add a new data source in Grafana and use the following URL:
 
+in case of running server locally:
```
academo, line 55: these instructions are no clear. if you run the mock server locally or docker the localhost url won't work because docker can't access it.

It depends in reality of how you run grafana, and most likely you will run it using docker if you are working with this example.

I suggest simply removing this instructions that calls for using localhost and keep the host.docker.internal reference only

academo (approved):

PR #273, opened 2024-04-11
Scenes app example: Add scenes app LLM plugin integration demo example

nmarrs:
Update the create plugin template scenes app to include a example integrating the LLM plugin

TODO:

- [ ] Test locally -  I'm not sure how to, but will work on figuring this out when I am back in next week
- [ ] Update documentation
- [x] Fix npm dependency issues
- [ ] re-test 😬 

In meantime, with LLM plugin configured here are the demos that should be possible with these changes 🦖

As a side note it would be super cool if there was a way to somehow have a single github repo for demos vs needing to duplicate changes between this repo and `plugin-tools` 😬 (i.e. [see this PR with duplicate changes](https://github.com/grafana/plugin-tools/pull/873))

https://github.com/grafana/scenes-app-template/assets/22381771/36b21e2d-aa4c-4515-baea-661c0764058c

https://github.com/grafana/scenes-app-template/assets/22381771/c67bcf88-3e25-4558-9647-02b0929a2ec3

https://github.com/grafana/scenes-app-template/assets/22381771/efe08fd8-fe64-47f9-b2e9-d198b061d610

Related PR: https://github.com/grafana/plugin-tools/pull/873

PR #322, opened 2024-06-19
Chore: App with scenes fix drill down page title when space is present in room name

nmarrs:
This change is broken out from https://github.com/grafana/grafana-plugin-examples/pull/273

Before we were taking raw URI encoded string and presenting it to user making two word room names have a `%20` show up as a space. Now we decode the URI so it is human readable.

I noticed this while creating [this basic demo](https://drive.google.com/file/d/1C3PDt3an9rcKjJtGsTpOTyYGnmypogKi/view?usp=sharing) getting a scenes app up and running in a local Grafana instance.

Before
Screenshot 2024-01-22 at 2 55 05 PM

After
Screenshot 2024-01-22 at 3 22 45 PM


andresmgot (approved):
LGTM

PR #332, opened 2024-07-05
Add example schema and validation hooks

andresmgot:
Demonstrating that generating schemas for queries work and that you can add an AdmissionHandler for an external plugin. The types are not get exposed in the API until [this](https://github.com/grafana/grafana/blob/main/pkg/registry/apis/datasource/register.go#L134) is resolved.

PR #335, opened 2024-07-09
Example: Add frontend migration example

andresmgot:
Example of how a change in the API + migration code may look in a plugin.

Note that this assumes that the plugin uses a schema and a API version. See this [base PR](https://github.com/grafana/grafana-plugin-examples/pull/332).

In this case, I am deprecating the field `multiplier` of the QueryData and renamed it `multiply`. Now both properties are optional, to avoid breaking changes in the `apiVersion`.

Note that the query is automatically migrated when running a query _but_ it's necessary for the plugin to manually call `migrateQuery` in the `QueryEditor` to ensure that it's using the latest version of the Query schema.

This is a demonstration for https://github.com/grafana/grafana/pull/90231

### DIFF
--- a/examples/datasource-http-backend/pkg/kinds/query.go
+++ b/examples/datasource-http-backend/pkg/kinds/query.go
@@ -6,7 +6,10 @@ import (
 )
 
 type DataQuery struct {
-	Multiplier int `json:"multiplier"`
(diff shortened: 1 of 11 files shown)


academo:
We are going to remove many examples in favour of what's on our documentation. I'll be closing this PR

PR #340, opened 2024-07-11
Example of backend query migration

andresmgot:
Example of how the migration handler can be used for a backend plugin.


academo:
We are going to delete many examples in favour of what we already have in our documentation. I'll be closing this PR.

PR #343, opened 2024-07-12
Docs: Fix links broken by plugin-tools site reorg

josmperez:
Fix links broken by plugin-tools site reorg.


sympatheticmoose (changes requested):
A lot of these changes should be resolved by running `create-plugin update` which is done periodically. Given we have redirects in place I would prefer to wait until that has run then catch any remaining links rather than have to review 82 files 😅

josmperez:
> A lot of these changes should be resolved by running `create-plugin update` which is done periodically. Given we have redirects in place I would prefer to wait until that has run then catch any remaining links rather than have to review 82 files 😅

Sounds good. Let's wait a bit.

academo:
@josmperez 
I am curious to know if we didn't setup a redirect for https://grafana.com/developers/plugin-tools/create-a-plugin/extend-a-plugin/extend-configurations#extend-the-eslint-config

because many projects have these files with this links and people might follow them if they open those files and haven't run update (which is not very often)

sympatheticmoose:
@academo there's a redirect in the re-org PR https://github.com/grafana/plugin-tools/pull/993/files#diff-d8d70f81bb443af1e4aa5f4b0ce4ba1cc15800b67be29a2bcc66187fb3c10d5cR193-R194

academo:
I'll be closing this PR in an effort to clean up the repository.

PR #364, opened 2024-09-03
Add auto-approve and auto merge into automatic update action

Ukochka:
Add ability to auto approve PR created by grafana-plugin bot to be able to automerge


andresmgot (approved):

.github/workflows/automatic-update.yml
```diff
@@ -43,3 +44,9 @@ jobs:
 
             **Notes for reviewer:**
             This is an auto-generated PR which ran `npx -y @grafana/create-plugin@latest update`, updates plugin.json files, grafana-plugin-sdk-go and docker-compose files for each plugin example
+
+      - name: Auto approve PR
+        uses: hmarr/auto-approve-action@v3
+        with:
+          pull-request-number: ${{ steps.create_pr.outputs.pull-request-number }}
+          github-token: ${{ secrets.GITHUB_TOKEN }}
```
andresmgot, line 52: shouldn't this be `token: ${{ steps.generate_token.outputs.token }}` as above?

.github/workflows/automatic-update.yml
```diff
@@ -43,3 +44,9 @@ jobs:
 
             **Notes for reviewer:**
             This is an auto-generated PR which ran `npx -y @grafana/create-plugin@latest update`, updates plugin.json files, grafana-plugin-sdk-go and docker-compose files for each plugin example
+
+      - name: Auto approve PR
+        uses: hmarr/auto-approve-action@v3
```
jackw, line 49:
```suggestion
        uses: hmarr/auto-approve-action@v4
```

The [action readme](https://github.com/hmarr/auto-approve-action/tree/v4) suggests using v4 due to deprecated versions of node.

.github/workflows/automatic-update.yml
```diff
@@ -43,3 +44,9 @@ jobs:
 
             **Notes for reviewer:**
             This is an auto-generated PR which ran `npx -y @grafana/create-plugin@latest update`, updates plugin.json files, grafana-plugin-sdk-go and docker-compose files for each plugin example
+
+      - name: Auto approve PR
```
jackw, line 48: Should this live in the same workflow or in a workflow that runs on `pull_request`? 🤔

jackw (commented):
I'm probably missing something but how does this auto merge the PR?

tolzhabayev:
I think you can skip the auto approve completely, as it doesn't make much of a difference.

SCR-20240905-kdwr

The bot user has the power to skip approvals and just merge it directly once the checks are green.

You would probably need something like https://github.com/re-actors/alls-green to ensure checks are actually green before merging.

A couple of examples on how to do it: https://jhale.dev/posts/auto-merging-prs/

.github/workflows/automatic-update.yml
Ukochka: no, it should not. Because if we are using the same token it thinks that it approves PR from itself. And it is not possible

Ukochka:
@jackw Auto-merge option is already in github and it is enabled for this repo. When all the requirements are there - it will be automatically merged. I thought we need only approval, but I see that we actually need something else instead. @tolzhabayev shared the example

tolzhabayev:
I think auto merge is not actually auto merge but an option you have to click on while the CI pipeline is running so that it will be merged once its done. But if you do not "click the button" it will not actually merge anything - e.g. this PR is not automatically merged although approved

Ukochka:
@tolzhabayev  yeah, that's why I was thinking auto-approve will help, but seems like no and we need to use auto-merge action to do so

jackw:
It seems we already have an [action in another repo](https://github.com/grafana/security-github-actions/blob/main/.github/workflows/dependabot-automerge.yaml) that does this work?

.github/workflows/automatic-update.yml
Ukochka: I am not sure, it is necessary. Is it?

Ukochka:
@tolzhabayev @jackw Hey guys, I have added merge part to this github actions. And I left approval part just to be sure. I think all-green is not needed here, because auto-merge anyway checks if everything is passed as I understand

mckn (commented):
Great work on this! Looks good 🙌🏻 

With that said I'm not sure I think this feature is a good idea. Isn't it enough magic to get an PR with a updated create-plugin repository. If I would have this running in my repository I would disable this feature because I would like to be in control of when things like this gets merged.

jackw:
@mckn I'm under the impression this action is intended only for the examples repo and isn't for plugin developers to use in their repos.

.github/workflows/automatic-update.yml
```diff
@@ -43,3 +44,15 @@ jobs:
...
+        uses: hmarr/auto-approve-action@v4
+        with:
+          pull-request-number: ${{ steps.create_pr.outputs.pull-request-number }}
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Enable auto-merge for PR
+        run: gh pr merge --auto --squash "$PR_URL"
```
jackw, line 55: Where does `$PR_URL` come from?

Does it make sense to enable auto-merge for the PR before approving it?

.github/workflows/automatic-update.yml
```diff
@@ -43,3 +44,15 @@ jobs:
 
             **Notes for reviewer:**
             This is an auto-generated PR which ran `npx -y @grafana/create-plugin@latest update`, updates plugin.json files, grafana-plugin-sdk-go and docker-compose files for each plugin example
+
+      - name: Auto approve PR
+        uses: hmarr/auto-approve-action@v4
```
jackw, line 49: Do we need a 3rd party action to do this? You're using the gh cli to auto merge the PR could we do the same with:

```
run: gh pr review ${{ steps.create_pr.outputs.pull-request-url }} --approve -b "Approving update"
```

.github/workflows/automatic-update.yml
jackw: I don't think it's necessary, but might be a little tidier having an action that runs with the `pull_request` event.

.github/workflows/automatic-update.yml
jackw: Ah ok I'm blind - the PR_URL is in the env below. 👍

tolzhabayev (approved):
Alright let's see this in action

jackw (approved):
LGTM! 🚀

tolzhabayev:
@Ukochka fixed the CLA check https://github.com/grafana/grafana-plugin-examples/pull/364#issuecomment-2334119860

PR #391, opened 2024-10-07
App with extension point: Fix failing e2e test

jackw:
This PR updates the app with extension point example to work with the latest canary version of Grafana by:

- including the extension info in the plugin.json so they can be registered.
- using the `usePluginLinks` hook to register the extension point so the app re-renders when the extensions are added to the registry.


marefr (approved):
LGTM

jackw:
Hmm... it's breaking the "expected version" tests now. I'll continue with it tomorrow.

leventebalogh (approved):
🚀

PR #396, opened 2024-10-11
chore: bumping go version 1.22

s4kh:
https://github.com/grafana/grafana-plugin-sdk-go/issues/1107

### Tests
Integration tests are passing.

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -92,7 +92,7 @@ jobs:
 
       - uses: actions/setup-go@v5
         with:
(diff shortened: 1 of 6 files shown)


andresmgot:
we probably need to update the version in the GH action too: https://github.com/grafana/grafana-plugin-examples/actions/runs/11293635621/workflow#L95 (to run tests with the new version)

marefr (approved):
LGTM. Think that the toolchain directives in go.mod can be removed/doesn't do any difference with these changes, but not sure

PR #398, opened 2024-10-14
Update Playwright to `v1.48.0`

leventebalogh:
@jackw spotted that our integration tests started failing due to the github runners using Ubuntu `24.04` instead of `22.04`. Turns out [we need to bump Playwright to support Ubuntu `24.04`](https://github.com/microsoft/playwright/issues/30368).


jackw (approved):
Awesome! 🚀

PR #400, opened 2024-10-15
datasource-http-backend: Add query definition

andresmgot:
Example of how a plugin can define query types and associated JSON schemas and examples.

The main definition is in `query.go`. The JSON files are automatically generated running `query_test.go`.


sunker (approved):
🚀

PR #401, opened 2024-10-15
datasource-http-backend: Add admission handler

andresmgot:
Example of how a plugin can define an `AdmissionHandler`. This is used to validate that given `DatasourceSettings` are valid.


academo:
I'll be closing this PR in an effort to clean up the repository. Feel free to reach out and open it again if you plan to work on it but we'll be removing most of these examples to favour our documentation.

PR #409, opened 2024-10-25
Fix panel test

sunker:
Having to look at grafana versions like this is not ideal. Might have to add an abstraction in plugin-e2e that bridges between versions when selecting panel options.


tolzhabayev (approved):

examples/panel-basic/tests/panel.spec.ts
```diff
@@ -14,16 +15,20 @@ test('should display series counter when "Show series counter" option is enabled
...
   await expect(page.getByTestId('simple-panel-series-counter')).not.toBeVisible();
-  const showSeriesSwitch = panelEditPage
-    .getByGrafanaSelector(selectors.components.PanelEditor.OptionsPane.fieldLabel('Basic Panel Show series counter'))
-    .getByLabel('Toggle switch');
-  await showSeriesSwitch.click();
+  const seriesCounterLabel = panelEditPage.getByGrafanaSelector(
+    selectors.components.PanelEditor.OptionsPane.fieldLabel('Basic Panel Show series counter')
+  );
+  const switchField = semver.gte(grafanaVersion, '11.4.0')
+    ? seriesCounterLabel.getByRole('switch')
```
andresmgot, line 30: I _think_ that `.getByRole('checkbox')` should work for both versions (checking these [docs](https://playwright.dev/docs/locators#locate-by-role)).

Also, another thing that you could try is to use a regex with `getByLabel`, like `.getByLabel(/.*/);`. In my head, that should select any label 🙃

sunker: The InlineSwitch is tricky when used in the panel options. One cannot target the checkbox as it's (weirdly) only the label that is clickable and not the checkout (see this issue for more details). Also since there are two labels rendered within Show series counter option, when cannot use `getByLabel(/.*/)` since it would return multiple labels (not possible to click multiple items in strict mode).

PR #484, opened 2025-03-07
Make plugin IDs match for `datasource-http-backend` example

wbrowne:


briangann (approved):
lgtm, the test failures are not related to these changes

examples/datasource-http-backend/src/plugin.json
```diff
@@ -2,7 +2,7 @@
   "$schema": "https://raw.githubusercontent.com/grafana/grafana/main/docs/sources/developers/plugins/plugin.schema.json",
   "type": "datasource",
   "name": "Datasource Http Backend",
-  "id": "example-httpbackend-datasource",
+  "id": "myorg-httpbackend-datasource",
```
andresmgot, line 5: sorry, this breaks some of my devs dashboards and other stuff, can we keep the same plugin ID? (I know it's incosistent between examples though).

wbrowne: We can sure but I was thinking it might be a good idea to make the IDs consistent at some point. Considering this is dev only, I figured breaking changes would be acceptable 🤷

andresmgot: yeah, not a big deal but it'd be still incosistent, some of them use `grafana` as the org, others `myorg` and others `example` 🤷

wbrowne: No that's what I meant 🙈 https://github.com/grafana/grafana-plugin-examples/pull/485

PR #485, opened 2025-03-07
Standardize org convention to avoid confusion

wbrowne:
I think consistency across the examples helps, and I believe having `"grafana"` as the the org can be confusing when people are unsure if/what should change. I chose `myorg` since we already have the majority using that already.


academo:
I'll be closing this PR in an effort to clean up the repository. Feel free to reach out and open it again if you plan to work on it.

PR #487, opened 2025-03-10
Revert "Make plugin IDs match for `datasource-http-backend` example"

wbrowne:
Reverts grafana/grafana-plugin-examples#484


andresmgot (approved):
thanks!

PR #493, opened 2025-03-20
Lapodocs: Trigger docs workflow on merge

academo:
Adds a github workflow to trigger grafana/plugin-tools when a change to a ts,tsx or go file is merged

### DIFF
--- a/.github/workflows/notify-plugin-tools.yml
+++ b/.github/workflows/notify-plugin-tools.yml
@@ -19,8 +19,8 @@ jobs:
         id: check_files
         run: |
           CHANGED_FILES=$(git diff --name-only ${{ github.event.pull_request.base.sha }} ${{ github.event.pull_request.head.sha }})
-          echo "$CHANGED_FILES" > changed_files.txt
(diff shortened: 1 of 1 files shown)


.github/workflows/notify-plugin-tools.yml
```diff
@@ -0,0 +1,37 @@
...
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Check for relevant file changes
+        id: check_files
+        run: |
+          CHANGED_FILES=$(git diff --name-only ${{ github.event.pull_request.base.sha }} ${{ github.event.pull_request.head.sha }})
+          echo "$CHANGED_FILES" > changed_files.txt
+          if grep -i -E "\.(ts|tsx|go)$" changed_files.txt; then
```
xnyo, line 23: Nit: To avoid writing directly on the repo I would change this to:

```suggestion
          echo "$CHANGED_FILES" > /tmp/changed_files.txt
          if grep -i -E "\.(ts|tsx|go)$" /tmp/changed_files.txt; then
```

xnyo (approved):
LGTM! Just one small nit:

PR #503, opened 2025-04-07
Panel: removes deprecated componets

### LINKED ISSUE #502: Replace TimeSeries, TooltipPlugin and ZoomPlugin
This [PR](https://github.com/grafana/grafana/pull/103490) removes a bunch of stuff from `@grafana/ui` we should make sure that our examples are up to date.

hugohaggmark:
This [PR ](https://github.com/grafana/grafana/pull/103490) will remove `TimeSeries`, `TooltipPlugin` and `ZoomPlugin` etc from `@grafana/ui`. This PR replaces them with an `Table` example until there is a valid migration path.

Fixes: #502

### DIFF
--- a/examples/panel-basic/README.md
+++ b/examples/panel-basic/README.md
@@ -2,9 +2,9 @@
 
 This example provides a template for how to build a basic panel plugin. The definition for the panel is provided in `/src/components/SimplePanel/SimplePanel.tsx`.
 
-The plugin uses `TimeSeries` from the `grafana-ui` package to build a graph with the properties passed to the panel. The plugin also allows a tooltip to be shown when the user hovers over a visualization.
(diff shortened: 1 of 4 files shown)


andresmgot (commented):
so if I read the thread correctly, the idea is to keep `TimeSeries` and the tooltip component?

In any case, the README of this example would need to be updated too.

hugohaggmark:
> so if I read the thread correctly, the idea is to keep `TimeSeries` and the tooltip component?
> 
> In any case, the README of this example would need to be updated too.

Awesome catch, I missed that 👍

andresmgot (commented):
Should we wait until we know if we can keep using `TimeSeries`?

PR #506, opened 2025-04-11
Chore: Ignore author field

tolzhabayev:
From the `peter-evans/create-pull-request` docs: 

https://github.com/peter-evans/create-pull-request/blob/a7b20e1da215b3ef3ccddb48ff65120256ed6226/docs/concepts-guidelines.md?plain=1#L378

> When setting `sign-commits: true` the action will ignore the `committer` and `author` inputs.

This PR comments out the `author` field to make sure only the fields that are used are actually set - less overall confusion.

### DIFF
--- a/.github/workflows/automatic-update.yml
+++ b/.github/workflows/automatic-update.yml
@@ -40,7 +40,7 @@ jobs:
         with:
           token: ${{ steps.generate_token.outputs.token }}
           branch: new-update-plugin-examples
-          #  Author field is ignored when commit-message is set to true
+          #  Author field is ignored when sign-commits is set to true
           # author: github-actions[bot] <41898282+github-actions[bot]@users.noreply.github.com>
           delete-branch: true
           commit-message: "chore: update configuration to latest version"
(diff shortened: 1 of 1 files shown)


.github/workflows/automatic-update.yml
```diff
@@ -40,7 +40,8 @@ jobs:
         with:
           token: ${{ steps.generate_token.outputs.token }}
           branch: new-update-plugin-examples
-          author: github-actions[bot] <41898282+github-actions[bot]@users.noreply.github.com>
+          #  Author field is ignored when commit-message is set to true
```
wbrowne, line 43: >commit-message is set to true

Should it be when `commit-message is set` or `commit-message is not empty'?

tolzhabayev: it's neither or, it's actually:

> when sign-commits is set to true

I don't know why I wrote commit-message before... fixed via https://github.com/grafana/grafana-plugin-examples/pull/506/commits/7d27e004aed1e151b387745768e317c03e7596a1

wbrowne (approved):

xnyo (approved):

PR #549, opened 2025-06-13
Chore: Harden github token permissions

oshirohugo:
This PR gives the github token used in CI only the minimum required permissions.
Part of https://github.com/grafana/grafana-community-team/issues/429

PR #550, opened 2025-06-13
Chore: Fix zizmor issues

oshirohugo:
Fix zizmor issues
https://github.com/grafana/grafana-plugin-examples/pull/549#issuecomment-2969623945

### DIFF
--- a/examples/app-basic/.github/workflows/ci.yml
+++ b/examples/app-basic/.github/workflows/ci.yml
@@ -64,14 +64,14 @@ jobs:
 
       - name: Test backend
         if: steps.check-for-backend.outputs.has-backend == 'true'
-        uses: magefile/mage-action@6f50bbb8ea47d56e62dee92392788acbc8192d0b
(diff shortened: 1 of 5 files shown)


examples/app-basic/.github/workflows/ci.yml
```diff
@@ -62,14 +64,14 @@ jobs:
 
       - name: Test backend
         if: steps.check-for-backend.outputs.has-backend == 'true'
-        uses: magefile/mage-action@v3
+        uses: magefile/mage-action@6f50bbb8ea47d56e62dee92392788acbc8192d0b
```
xnyo, line 67: You should add a comment next to each hash with the git tag for the action. This will allow dependabot to update it. Like so:

```suggestion
        uses: magefile/mage-action@6f50bbb8ea47d56e62dee92392788acbc8192d0b # v3
```

xnyo (approved):
LGTM great work! :rocket:

PR #555, opened 2025-06-16
Chore: Use plugins platform bot for examples update integration

academo:
Changes the workflow to use the plugins platform bot instead so PR checks can run and pass


jackw (approved):
LGTM! 🚀

PR #590, opened 2025-09-23
Cleanup ci workflow

sunker:
Historically, when this repo depended on `grafana/e2e`, the CI workflow needed to determine which version of the Grafana packages to install before running e2e tests against a canary Grafana build. With the latest version of `plugin-e2e`, this is no longer necessary, as it works across all Grafana versions.

This PR removes that legacy step from the CI workflow and cleans up the process overall. It also simplifies the script used to retrieve the latest `grafana-dev` image tag. The script is copied from the [`plugin-actions` repo](https://github.com/grafana/plugin-actions/pull/118). In the future, we may want to move this logic into a composite action for reuse, but for now this approach is sufficient.


jackw (approved):
LGTM! 🚀

PR #653, opened 2026-04-06
Fix e2e tests failing due to new splashScreen

andresmgot:


andresmgot (commented):
not sure why this didn't work 🤔

sunker:
> not sure why this didn't work 🤔

I guess the env variable is not correctly formatted? Either way fixing it here only temporarily solves the problem as these plugins gets regenerated from create-plugin once a week. Since yesterday, plugin-e2e handles this (sets the spashScreen toggle to false by default) so I think you can close this PR.

andresmgot:
Great, thank you!

PR #724, opened 2026-07-27
ci(renovate): post gate verdict from a separate step

tolzhabayev:
The renovate review now ends with a verdict line and a separate step posts the approval or the needs-human comment. Runs that produce no verdict fail, so the sweeper retries them. Also lets the review read release notes (gh release view, WebFetch) and raises the turn cap to 50.

The examples-refresh job keeps posting its own review (it approves as github-actions since the bot authors those PRs), so it gets the same outcome check as a separate step plus the higher turn cap.


leventebalogh (approved):

tolzhabayev:
Superseded by #725, which rewrites this file for the scheduled trigger and already carries the verdict split. The pieces this PR added to the examples job (turn limit 50, persist-credentials, and the step that fails a run which posted nothing) are folded into the new examples-refresh.yml there.

